PR #12, opened 2020-07-03
Add a pipeline for deploying production releases

TurpIF:
The new deployment workflow use the GitHub releases as a trigger. For 
the moment, creating the release is a manual step (consisting of 
copy/paste of info in the slack notif). We could try to work later on it
to make the release fully automated.

JIRA: EE-1166

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+name: Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    if: "!github.event.release.prerelease"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Android SDK
+        uses: malinskiy/action-android/install-sdk@release/0.0.7
+
+      - name: Deploy artifacts and notify on Slack
+        env:
+          AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}
+        run: ./gradlew publishReleasePublicationToAzureRepository \
+          sendReleaseDeployedToAzureMessageToSlack \
+          -PisRelease=true
--- a/buildSrc/src/main/java/Slack.kt
+++ b/buildSrc/src/main/java/Slack.kt
@@ -14,6 +14,7 @@
  *    limitations under the License.
  */
 
+import com.slack.api.model.block.composition.BlockCompositions.markdownText
 import org.gradle.api.Project
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.kotlin.dsl.withType
@@ -33,7 +34,6 @@ private fun Project.addSlackDeploymentMessage(publication: MavenPublication, rep
   val teamChannel = "#pub-sdk-private"
   val rcChannel = "#pub-sdk-release-candidates"
   val confluenceSpaceUrl = "https://go.crto.in/publisher-sdk-bugfest"
-  val gerritProjectBranchesUrl = "https://go.crto.in/publisher-sdk-android-branches"
 
   slack {
     messages {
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
+name: Release a new version
```
vguerci, line 1: `Release` is enough

.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    if: "!github.event.release.prerelease"
```
vguerci, line 9: Don't you want to release RC as pre-release?

.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Android SDK
+        uses: malinskiy/action-android/install-sdk@release/0.0.7
+
+      - name: Deploy artifacts and notify on Slack
+        env:
+          AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}
```
vguerci, line 20: Still using azure? Thought we were moving out

buildSrc/src/main/java/Slack.kt
```diff
@@ -61,23 +61,41 @@ private fun Project.addSlackDeploymentMessage(publication: MavenPublication, rep
...
-- Go on <$gerritProjectBranchesUrl|Gerrit> and create the `v${sdkVersion()}` branch on this commit SHA-1 (only for new version)
+- Run `$gitCommand`
 - Share this message on $teamChannel
 *Validate the RC*
 - Install the <${testAppUrl(publication.version)}|TestApp APK>
 - Go on <$confluenceSpaceUrl/Bugfest+Android+${publication.version}|Bugfest page> and execute tests
 *Release the RC*
-- Go on <https://go.crto.in/publisher-sdk-prod-deployment|Jenkins deploy job> and insert this commit SHA-1
+- Create a <https://github.com/criteo/android-publisher-sdk/releases/new|new GitHub release> and insert:
+  - Tag version: `v${sdkVersion()}`
```
vguerci, line 95: Creating a Release on GitHub automatically creates tag if missing

buildSrc/src/main/java/Slack.kt
```diff
@@ -61,23 +61,41 @@ private fun Project.addSlackDeploymentMessage(publication: MavenPublication, rep
...
+  - Target: `v${sdkVersion()}`
+  - Release title: `Release v${sdkVersion()}`
```
vguerci, line 97: Release title is generally just the version `v${sdkVersion()}`

buildSrc/src/main/java/azure/AzureBlobStorage.kt
```diff
@@ -103,9 +103,9 @@ class AzureBlobStorage(private val containerName: String) {
     val blob = blobContainer.getBlockBlobReference(blobName)
 
     println("Upload $path to $blobName")
-    Files.newInputStream(path).use {
-      blob.upload(it, Files.size(path))
-    }
+//    Files.newInputStream(path).use {
```
vguerci, line 106: I don't get this, could you please elaborate?

buildSrc/src/main/java/Slack.kt
```diff
@@ -61,23 +61,41 @@ private fun Project.addSlackDeploymentMessage(publication: MavenPublication, rep
...
+- Run `$gitCommand`
 - Share this message on $teamChannel
 *Validate the RC*
 - Install the <${testAppUrl(publication.version)}|TestApp APK>
 - Go on <$confluenceSpaceUrl/Bugfest+Android+${publication.version}|Bugfest page> and execute tests
 *Release the RC*
-- Go on <https://go.crto.in/publisher-sdk-prod-deployment|Jenkins deploy job> and insert this commit SHA-1
+- Create a <https://github.com/criteo/android-publisher-sdk/releases/new|new GitHub release> and insert:
+  - Tag version: `v${sdkVersion()}`
+  - Target: `v${sdkVersion()}`
```
vguerci, line 96: Target should be left as default, `main`

.github/workflows/release.yml
TurpIF: Done

.github/workflows/release.yml
TurpIF: Yep that would be great. For the moment I deactivate the Azure deployment in case of prerelease to be sure that, in the future when we'll work on RC as prerelease, there is no risk to trigger a real release

.github/workflows/release.yml
TurpIF: Yep, we need first to ask Bintray to sync our artifacts on JCenter. Once sync is automated, we can move forward on JCenter and drop Azure. (See EE-1167)

buildSrc/src/main/java/Slack.kt
TurpIF: Done and merge three lines in one

buildSrc/src/main/java/azure/AzureBlobStorage.kt
TurpIF: One thing that we should keep in mind with PR: commit messages are not really visible :)
In this case, I temporary deactivate the publication on Azure, so we can test the new release process and check if there is no issue.
I added a comment in the code to explain this. Of course, I will reactivate this after the test.

buildSrc/src/main/java/Slack.kt
TurpIF: On Android process, to let main be updated while doing a release, we're always making a new branch for the new version. Hence we should target the new branch instead of the master.
We'll see if moving on GA will make the release process even smoother than we won't need to have such branch.

vguerci: I still do not understand why creating a dead version branch where a tag is enough ¯\\\_(ツ)_/¯

TurpIF: I case of buggy RC, fixes can be applied on the version branch. In the meantime, master can live independently.

PR #13, opened 2020-07-03
Run instrumentation tests in Github Action and prepare tests to be run

TurpIF:
In parallel of Java tests, instrumentation tests should be run on a
single emulator. If emulator crashes, logcat should be available as a
temporary build artifact. If test fails, test report should be available
as a temporary build artifact

As this is totally new, some tests are expected to fail. Easy-to-fix tests might be fixed. Others might be ignored for now a fixed later in EE-1180.

JIRA: EE-1165

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ on:
       - main
 
 jobs:
-  build:
+  java-tests:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -29,12 +29,44 @@ jobs:
           path: "**/TEST-*.xml"
           testSrcPath: "**/test/"
 
+  android-tests:
```
vguerci, line 32: #L12 should be renamed from `build` to `tests` or `java-tests`

.github/workflows/main.yml
```diff
@@ -29,12 +29,44 @@ jobs:
...
+          access-token: ${{ secrets.GITHUB_TOKEN }}
+          path: "**/TEST-*.xml"
+          testSrcPath: "**/test/"
+
   deploy-development-artifacts:
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/main'
 
     needs:
       - build
```
vguerci, line 68: Same, `java-tests`

TurpIF: done

.github/workflows/main.yml
TurpIF: done, renamed to java-tests

TurpIF:
Without the first commit on CSM, the tests crashes in the middle of the run. It is because, even when we recovered from Tape queue corruption, an exception was thrown in tests. And sometimes the exception was thrown outside the test thread, which caused the whole test suit to stop.

The last commit on native, is a quick fix for functional tests on Advanced Native (We were aware of this issue but as it was green on our laptop, we didn't take times for this). If you prefer, I can remove this and investigate more in EE-1180.

vguerci:
My concern is that I've started this review which was about adding Android Tests, and that I now see unrelated commits appearing in the PR, changing the scope of the review. In a former project we used to Squash & Merge PR, having in the end a linear commit history like on Gerrit. But that implied a PR to be like a commit, atomic, maybe I'm biased because of this.

vguerci:
In any cases, commits such as fixing CSM are unrelated from the PR description, so I think you should change PR title/description.

TurpIF:
When OK to merge, I will remove the expectation on the old `build` job, and and expectations on the new `java-tests` and `android-test`.

Amokrane:
@TurpIF Agreed with your approach. Let's make it mandatory for merging first and see how far we can go with that.

PR #18, opened 2020-07-07
Move in progress changes in a dedicated section of the CHANGELOG.md

TurpIF:
This is to avoid having publisher looking at the changelog and expecting
features/bugs in a version that is not yet released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Version 3.7.1
+# Next version
 
 - Features
   - Provide legal privacy text for native in `CriteoNativeAd#getLegalText`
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Version 3.7.1
+# Next version
```
Amokrane, line 1: I recommend bumping to 3.8.0 instead since it's a feature change.

TurpIF:
What about the risk that publisher

TurpIF:
You mean add a 3.8.0 section in the changelog ?
What about the risk that some publishers may see this and then expect that 3.8.0 is already released ?

PR #19, opened 2020-07-07
Deploy the test app in a dedicated package on Bintray

TurpIF:
JCenter only accept packages representing library with .jar or .aar
The test app is not in this case, and there is no meaning to sync it on
JCenter anyway. Hence, we push it on Bintray but in a dedicated package.
So the SDK and the test utils can be sync with JCenter.

This also remove the declaration of the specific Bintray repo of Criteo.
As mediation adapters are now sync with JCenter, this is not needed
anymore.

JIRA: EE-1167

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,6 +1,3 @@
-import Deps.Criteo.Mediation.adMobAdapter
-import Deps.Criteo.Mediation.moPubAdapter
-
 /*
  *    Copyright 2020 Criteo
  *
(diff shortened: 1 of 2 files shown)


app/build.gradle.kts
```diff
@@ -60,7 +60,14 @@ addPublication("Apk") {
     }
 }
 
-addBintrayRepository()
+addBintrayRepository {
+    // JCenter only accept packages representing library with .jar or .aar
```
Amokrane, line 64: nit:  JCenter only accepts packages representing libraries with .jar or .aar

TurpIF: done

PR #25, opened 2020-07-10
Store cache between CI runs

TurpIF:
The Gradle files does not change that much. When there is no change, the
gradle cache is cached between runs on the CI. It is expected that this
speed up the build time.

It would also be nice to cache the Android SDK because this can take a while. But the action we're currently using does not seem to fully accept recycle SDK when (https://github.com/Malinskiy/action-android/issues/26).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Store Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: ${{ runner.os }}-gradle-
+
       - name: Set up Android SDK
         uses: malinskiy/action-android/install-sdk@release/0.0.7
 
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -40,6 +54,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Store Gradle cache
```
Amokrane, line 57: Duplicate job? (already defined above)

.github/workflows/main.yml
```diff
@@ -15,9 +15,23 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: ${{ runner.os }}-gradle-
+
       - name: Set up Android SDK
         uses: malinskiy/action-android/install-sdk@release/0.0.7
 
+      - name: Store Android SDK
```
Amokrane, line 28: Could you add a comment that says that this won't work until https://github.com/Malinskiy/action-android/issues/26 is addressed?

.github/workflows/main.yml
TurpIF: This one is for the andorid-tests job, while the one above is for java-tests

.github/workflows/main.yml
TurpIF: This half-work:
- This step cannot be put before the "Set up Android ADK" because of the issue => tools, platform-tools, licenses are downloaded
- By being there, this step remove all the build-tools setup done by gradle

I will add a comment explaining this.

TurpIF: Done

PR #27, opened 2020-07-10
Log DFP failedToLoad error in tests

Amokrane:
This will add some context which would help us understand why dfp tests
fail on GitHub Actions.

### DIFF
--- a/publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/integration/DfpHeaderBiddingFunctionalTest.java
+++ b/publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/integration/DfpHeaderBiddingFunctionalTest.java
@@ -32,10 +32,13 @@ import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.when;
 
 import android.os.Bundle;
+import android.util.Log;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/integration/DfpHeaderBiddingFunctionalTest.java
```diff
@@ -597,9 +599,10 @@ public void onAdLoaded() {
 
       @Override
       public void onAdFailedToLoad(int i) {
+        Log.w(TAG, "onAdFailedToLoad, for reason: " + i +
```
TurpIF, line 602: Nit: you can also use the Logger class (see https://github.com/criteo/android-publisher-sdk/blob/7c1b0b91f8a9dc78ba25b8850d51d5d51d244516/publisher-sdk/src/main/java/com/criteo/publisher/bid/LoggingBidLifecycleListener.java#L32) that handle %s, %d, ... formatting with varargs.

Amokrane: Done

PR #33, opened 2020-07-17
Improve observability when SDK fail to get advertising info

TurpIF:
The methods that are used in AdvertisingInfo are in the compile classpath and are expected to be
provided in the runtime classpath by publishers as explained in the public documentation. Hence reflection can be removed.

Logging in case of error is improved:
- if publishers didn't brings the good dependency, they are notified to
- sometimes, in tests, it is impossible to read the advertising ID and
there is no clue of why. The stacktrace will now be printed in logcat.

The DeviceUtil class does not bring any value to the AdvertisingInfo
class. Worse, it steals the responsibility to use the dummy ID in case
of advertiser info limited.


TurpIF:
Ok, if you need to come back on this later, do not hesitate to tell me

PR #37, opened 2020-07-21
Use integration registry instead of const profile ID

TurpIF:
- [x] CDB network call use integration registry
- [x] Remote config network call use integration registry
- [x] CSM is fully derived from bidding

JIRA: EE-1198

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/csm/MetricSendingQueueConsumer.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/csm/MetricSendingQueueConsumer.java
@@ -17,17 +17,9 @@
 package com.criteo.publisher.csm;
 
 import androidx.annotation.NonNull;
-import com.criteo.publisher.SafeRunnable;
-import com.criteo.publisher.integration.Integration;
 import com.criteo.publisher.model.Config;
 import com.criteo.publisher.network.PubSdkApi;
 import com.criteo.publisher.util.BuildConfigWrapper;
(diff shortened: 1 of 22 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/csm/MetricSendingQueueConsumer.java
```diff
@@ -100,22 +105,50 @@ public void runSafely() throws IOException {
...
+      for (Entry<Integer, Collection<Metric>> entry : metricsPerProfile.entrySet()) {
+        int profileId = entry.getKey();
+        Collection<Metric> metricsForProfile = entry.getValue();
+        MetricRequest metricRequest = MetricRequest.create(
+            metricsForProfile,
+            sdkVersion,
+            profileId
+        );
+        metricRequests.put(metricRequest, metricsForProfile);
+      }
```
vguerci: Would it make sense that `MetricRequest` keep & expose its underlying metrics, as it takes `metricsForProfile` as input, rather than this `Map<MetricRequest, Collection<Metric>`?

TurpIF: The `MetricRequest` is a DTO object autogenerated from an interface and JSON schema is based directly on its properties. I would avoid putting too much logic in this class.

Also, here the `metricForProfile` is kept only to handle the partial rollback.

PR #40, opened 2020-07-24
Use fixed version for the upload-artifact GA

TurpIF:
This GA is now using "main" as default branch. Using the master one
yield this warning:

The "master" branch is no longer the default branch name for
actions/upload-artifact
Please pin to a specific version or use "main".

To stay out of any change a fixed version is used instead.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
           access-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Upload JUnit report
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           access-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Upload JUnit report
-        uses: actions/upload-artifact@master
+        uses: actions/upload-artifact@main
```
vguerci, line 46: Please target @v2

PR #45, opened 2020-07-29
Setup Gordon Gradle plugin to retry failing tests

TurpIF:
This will retry failing tests up to 3 times. This plugin wrap the
connected Android tests and its main features are:
- Retrying failing tests
- Generating JUnit XML reports (with syserr for recovered flaky tests)
- Generating HTML reports with failing/recovering/passing tests
- Printing ongoing tests in stdout without --info on Gradle logger

Why do we need it ?
Because your tests are so flaky that the devs have to use Head&Shoulders
dandruff shampoo</Gordon Ramsay>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,7 @@ jobs:
         with:
           api: 29
           tag: google_apis
-          cmd: ./gradlew :publisher-sdk-tests:connectedCheck --info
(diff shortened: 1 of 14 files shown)


TurpIF:
This is blocked because of https://github.com/Banno/Gordon/issues/46 and https://github.com/Banno/Gordon/issues/47

PR #52, opened 2020-08-04
Provide ZoneId in CSM

TurpIF:
JIRA: EE-1124


publisher-sdk/src/main/java/com/criteo/publisher/csm/Metric.java
```diff
@@ -68,6 +68,14 @@
   @Nullable
   abstract String getRequestGroupId();
 
+  /**
+   * Zone ID that was mapped during bidding to the CDB slot response.
+   * <p>
+   * This should be present in case of valid bid.
```
Amokrane, line 74: should or must? https://www.ietf.org/rfc/rfc2119.txt

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponseSlot.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.criteo.publisher.DependencyProvider
+import com.criteo.publisher.annotation.OpenForTesting
+import com.criteo.publisher.model.nativeads.NativeAssets
+import com.criteo.publisher.util.URLUtil
+import com.google.gson.annotations.SerializedName
+import org.json.JSONObject
+import java.io.ByteArrayInputStream
+
+@OpenForTesting
+data class CdbResponseSlot(
```
Amokrane, line 29: Why was `Slot` renamed to `CdbResponseSlot` during the data class migration? Is there another `Slot` class?
I prefer `Slot` for simplicity.

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponseSlot.kt
```diff
@@ -0,0 +1,83 @@
...
+    @SerializedName("currency") val currency: String? = null,
+    @SerializedName("width") val width: Int = 0,
+    @SerializedName("height") val height: Int = 0,
+    @SerializedName("displayUrl") val displayUrl: String? = null,
+    @SerializedName("native") val nativeAssets: NativeAssets? = null,
+
+    /**
+     * The TTL in seconds for this bid response.
+     */
+    @SerializedName("ttl") var ttl: Int = 0,
```
Amokrane, line 43: I suggest renaming to `ttlInSeconds`.

publisher-sdk/src/main/java/com/criteo/publisher/csm/Metric.java
TurpIF: Normally it is "must". But we do not control the TLA and this information comes from it.

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponseSlot.kt
TurpIF: The is the `CdbRequestSlot` too. This is to make the distinction.

PR #57, opened 2020-08-10
Don't allow non foreground processes to init Criteo

Amokrane:
Defensive action against cases where Criteo is initialized
in a background service. If such service is always restarted,
Criteo will initialize many times, causing many prefeteches.

JIRA: EE-648


publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/integration/CriteoFunctionalTest.java
```diff
@@ -169,6 +177,23 @@ public void init_GivenCpIdAppIdAndVersion_CallConfigWithThose() throws Exception
...
+    assertNotNull(criteo);
+    assertFalse(criteo instanceof DummyCriteo);
+  }
+
+  @Test
+  public void init_GivenCurrentForegroundProcess_DummyCriteoIsInitialized() throws Exception {
+    when(androidUtil.isCurrentProcessInForeground(application)).thenReturn(false);
+    Criteo criteo = givenInitializedCriteo();
+    waitForBids();
+    assertNull(criteo);
```
TurpIF, line 194: Shouldn't you check that dummy criteo is returned ?

publisher-sdk/src/main/java/com/criteo/publisher/Criteo.java
```diff
@@ -104,8 +105,11 @@ private static Criteo init(
           dependencyProvider.setCriteoPublisherId(criteoPublisherId);
 
           DeviceUtil deviceUtil = dependencyProvider.provideDeviceUtil();
+          AndroidUtil androidUtil = dependencyProvider.provideAndroidUtil();
 
-          if (deviceUtil.isVersionSupported()) {
+          if (!androidUtil.isCurrentProcessInForeground(application)) {
+            criteo = null;
```
TurpIF, line 111: This make this method nullable. Hence it is not safe anymore to do init+getinstance. Maybe could you use the dummy criteo instead ?

test-utils/src/main/java/com/criteo/publisher/mock/MockedDependenciesRule.java
```diff
@@ -224,9 +226,17 @@ public void resetAllDependencies() {
 
     setUpDependencyProvider();
     setUpCdbMock();
+    setupAndroidUtilMock();
     injectDependencies();
   }
 
+  private void setupAndroidUtilMock() {
+    AndroidUtil androidUtil = new AndroidUtil(dependencyProvider.provideApplication());
```
TurpIF, line 234: Instead of instantiating the AndroidUtil, why don't you ask the dependency provider for one ?

publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/integration/StandaloneFunctionalTest.java
```diff
@@ -108,6 +108,7 @@ public void setUp() throws Exception {
 
     api = spy(dependencyProvider.providePubSdkApi());
     when(dependencyProvider.providePubSdkApi()).thenReturn(api);
+    when(androidUtil.isCurrentProcessInForeground(any())).thenReturn(true);
```
TurpIF, line 111: Nit: you can transform the `@Mock` to `@SpyBean` to make it instantiated via the mocked dependency provider.
Honestly, I don't know if it work if you use `@MockBean`. As the annotations are processed after your code in `MockedDependencyProvider`, the annotations processor might not copy the previous stub registered.

publisher-sdk/src/main/java/com/criteo/publisher/util/AndroidUtil.java
```diff
@@ -57,4 +63,25 @@ public int dpToPixel(int dp) {
     return (int) Math.ceil(dp * context.getResources().getDisplayMetrics().density);
   }
 
+  /**
+   * Check whether the current process is in foreground.
+   *
+   * @return {@code true} if the current process is in foreground, {@code false} otherwise
+   */
+  public boolean isCurrentProcessInForeground(@NonNull Context context) {
```
TurpIF, line 71: Do you know a way to test this ? It would be great to cover such kind of low level code

publisher-sdk/src/main/java/com/criteo/publisher/Criteo.java
Amokrane: The dummy criteo instance wouldn't work. You could have a case where Criteo is initialized from a background service, then initialized properly from a foreground process. When this happens, the normal initialization won't happen as criteo would already be initialized and stuck on `DummyCriteo` until next process restart. If the service always call `Application#onCreate` first, then Criteo would never initialize properly.
We could make it work by changing the condition here: https://github.com/criteo/android-publisher-sdk/blob/main/publisher-sdk/src/main/java/com/criteo/publisher/Criteo.java#L100, to be: 

`if (criteo == null || criteo instanceof DummyCriteo)`

TurpIF: Oh ok, I wasn't aware of this. Ok to change the condition then. Please also add a comment explaining the why of such condition.
Moreover, could you add a test validating the scenario you described ?

publisher-sdk/src/main/java/com/criteo/publisher/util/AndroidUtil.java
Amokrane: I will try to cover this using https://developer.android.com/reference/androidx/test/rule/ServiceTestRule

Amokrane:
Abandoning this PR, after considering some nasty side effects that come from this strategy. In particular, this could lead to never initialize Criteo correctly if a service always causes the first Criteo initialization (Criteo will always be a Dummy instance). A rate limiter strategy will be used instead (PR incoming).

TurpIF:
Do not forget to remove the branch when you're finished with this work 😉

PR #58, opened 2020-08-11
Add an API to support server side bidding

Amokrane:


Amokrane:
@TurpIF  this is a draft to get your early feedbacks and check if I have missed anything important.

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_server_bidding)
+
+    val buttonBanner: Button = findViewById(R.id.buttonBanner)
+    buttonBanner.setOnClickListener {
+      val bannerViewContainer: LinearLayout = findViewById(R.id.bannerViewContainer)
+      criteoBannerView = CriteoBannerView(this.baseContext, BANNER_ADUNIT)
+      // todo: replace with a fake/demo ajs url
```
TurpIF, line 48: Demo AJS: https://rdi.eu.criteo.com/delivery/rtb/demo/ajs?zoneid=1417086&width=300&height=250&ibva=0

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+  private lateinit var criteoBannerView: CriteoBannerView
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_server_bidding)
+
+    val buttonBanner: Button = findViewById(R.id.buttonBanner)
+    buttonBanner.setOnClickListener {
+      val bannerViewContainer: LinearLayout = findViewById(R.id.bannerViewContainer)
+      criteoBannerView = CriteoBannerView(this.baseContext, BANNER_ADUNIT)
```
TurpIF, line 47: The banner_adunit is useless here.
This will make two use cases with a useless parameter: In-House and server-side bidding. What do you think about adding a new constructor without ad unit ? (and with proper comment)

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+      val criteoInterstitial = CriteoInterstitial(this.baseContext, INTERSTITIAL_ADUNIT)
+      criteoInterstitial.setCriteoInterstitialAdDisplayListener(object: CriteoInterstitialAdDisplayListener {
+        override fun onAdFailedToDisplay(error: CriteoErrorCode?) {
+        }
+
+        override fun onAdReadyToDisplay() {
+          criteoInterstitial.show()
+        }
+      })
+      // todo: replace with a fake/demo ajs url
```
TurpIF, line 64: This is already a demo AJS, isn't it ?

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+      val bannerViewContainer: LinearLayout = findViewById(R.id.bannerViewContainer)
+      criteoBannerView = CriteoBannerView(this.baseContext, BANNER_ADUNIT)
+      // todo: replace with a fake/demo ajs url
+      criteoBannerView.displayAd("https://ads.eu.criteo.com/delivery/r/ajs.php?did=5f316b617f6e80e2414aa4c068a9a400&u=%7CQnj3nJUtcJBmcy1yUzVZ0sJJVh1R2jYzX03kRoCOGY8%3D%7C&c1=bXzSPZ7XI4X4_eFA_ib-N7k6ayqfn4hlMBTatiYKxuTGVTBcI1VNzEbwEKS44St5EQ4YnwKmKigxwXfKiVfcgEOBtjH0Hh5NsGXtNLl4JkPP8mFeu4Apmc5mfClzCS8yYqxR30zVSw4xDDHw0Vm62BhPaN9RcfzdiS9AbB9nFqbL7um63XLc3suVOCz3emERkg9VaS20rlaR9500TX7BG-L7xYlhaVaZhWmPbvx2xJvLkjNqWiCBpWmejcw7e3my5aWP3cq1MgFY6znEwehZVuF-DflgEeIqGTqhxejhNNl4lZIVFZwtAf1FV1Wo-Kf9gBZejgIknyScKDfX3zha4sfV7hMVU34Btq6evxpZz4Kj9EPSNaPY1EXa3mfcmokLMF6EhZmbHBC_o7tTx5UkBznotTA1Q9ojbajl-oaObczVpHPVdWnjQWl6bPYU2oEoVbwnFR53gWoi_IZU1LME0A")
+      bannerViewContainer.addView(criteoBannerView)
+    }
+
+    val buttonInterstitial: Button = findViewById(R.id.buttonInterstitial)
+    buttonInterstitial.setOnClickListener {
+      val criteoInterstitial = CriteoInterstitial(this.baseContext, INTERSTITIAL_ADUNIT)
```
TurpIF, line 55: Same comment than above: ad unit is useless, what about adding a constructor without it ?

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+          criteoInterstitial.show()
+        }
+      })
+      // todo: replace with a fake/demo ajs url
+      criteoInterstitial.displayAd("https://rdi.eu.criteo.com/delivery/rtb/demo/ajs?zoneid=1417086&width=393&height=759&ibva=1&uaCap=5")
+    }
+  }
+
+  override fun onDestroy() {
+    super.onDestroy()
```
TurpIF, line 70: shouldn't you destroy the activity after destroying the view ?

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -78,6 +78,12 @@ public void loadAd() {
     }
   }
 
+  public void displayAd(@NonNull String signalData) {
+    // FIXME: call notifyFor? Assuming publishers would pass a listener during the main SDK setup
```
TurpIF, line 82: Could you explain more what you're expecting ?
This will call the onAdReceived of the listener. Semantically, ad is already received at this point.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -78,6 +78,12 @@ public void loadAd() {
     }
   }
 
+  public void displayAd(@NonNull String signalData) {
```
TurpIF, line 81: signalData does not seem legit here. Is it coming from openrtb or the server side bidding ?
I would propose adData, displayAdData, displayToken, displayData, ...

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitial.java
```diff
@@ -98,6 +98,11 @@ public void loadAd(@Nullable BidToken bidToken) {
     }
   }
 
+  public void displayAd(@NonNull String signalData) {
+    getOrCreateController().notifyFor(CriteoListenerCode.VALID);
```
TurpIF, line 102: Same than for banner.
Moreover, the display listener will take care of notifying publisher for success/failure of display

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
Amokrane: Indeed.

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
Amokrane: Done. Though for `In-House` why isn't the AdUnit needed? I see that we are passing the AdUnit here: https://github.com/criteo/android-publisher-sdk/blob/main/app/src/main/java/com/criteo/pubsdk_android/InHouseActivity.java#L110

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
Amokrane: Done. By the way, do you recall the story behind the unused `context`?

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: It's terminology coming from the partner we are working with. Ok to use `displayData` instead.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitial.java
Amokrane: Removed. It doesn't make sense indeed.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: Removed. It doesn't make sense indeed.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitialEventController.java
```diff
@@ -124,9 +124,10 @@ void fetchCreativeAsync(@NonNull String displayUrl) {
   }
 
   public void show() {
-    if (!isAdLoaded()) {
-      return;
-    }
+    // FIXME: unccomment when done
```
Amokrane, line 127: oops.

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
TurpIF: For InHouse, the AdUnit is given to the `Criteo.getBidResponseFor(adUnit)`. Then the token is given to the banner view. So the ad unit given to the banner view's constructor is not really useful here.

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
TurpIF: The context was used before to open the interstitial activity. But now this is handled by the `InterstitialActivityHelper` class which gets its context from the `DependencyProvider`.
Btw, removing the context parameter might be added in the TODO list of breaking changes for next major version.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -61,6 +57,18 @@ public CriteoBannerView(@NonNull Context context, @Nullable BannerAdUnit bannerA
     this.criteo = criteo;
   }
 
+
+  public CriteoBannerView(@NonNull Context context, @Nullable BannerAdUnit bannerAdUnit) {
```
TurpIF, line 61: Nit:
Generally, public constructors are above internal ones.

app/src/main/java/com/criteo/pubsdk_android/ServerBiddingActivity.kt
Amokrane: Thanks, clear. Created EE-1249 for that.

TurpIF: EE-1229 😉

publisher-sdk/src/androidTest/java/com/criteo/publisher/CriteoBannerViewIntegrationTest.java
```diff
@@ -49,13 +50,22 @@ public void setup() throws Exception {
 
     runOnMainThreadAndWait(() -> {
       criteoBannerView = new CriteoBannerView(
-          InstrumentationRegistry.getContext(),
+          InstrumentationUtil.getApplication().getApplicationContext(),
```
TurpIF, line 53: Nit: as you have access to the mocked dependency provide, you might get a context by adding `@Inject private lateinit var context: Context` in the field section.

publisher-sdk/src/androidTest/java/com/criteo/publisher/CriteoBannerViewIntegrationTest.java
```diff
@@ -49,13 +50,22 @@ public void setup() throws Exception {
...
     });
 
     criteoBannerView.setCriteoBannerAdListener(criteoBannerAdListener);
   }
 
+
+  @Test
+  public void instantiate_GivenContext_InitializeCorrectly() throws Exception {
+    runOnMainThreadAndWait(() -> {
+      CriteoBannerView criteoBannerView = new CriteoBannerView(InstrumentationUtil.getApplication().getApplicationContext(), null);
```
TurpIF, line 64: Don't you want to test the constructor without the second parameter ?

publisher-sdk/src/androidTest/java/com/criteo/publisher/CriteoBannerViewIntegrationTest.java
```diff
@@ -49,13 +50,22 @@ public void setup() throws Exception {
...
+      assertThat(criteoBannerView.getContext()).isEqualTo(InstrumentationUtil.getApplication().getApplicationContext());
```
TurpIF, line 65: `InstrumentationUtil...` is used twice here, (three times with the one in the setup method. What do you think about saving it in a local val or in a field ?

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -49,6 +49,13 @@
   @Nullable
   private CriteoBannerEventController criteoBannerEventController;
 
+  /**
+   * Will be used for server side bidding and in-house auction
```
TurpIF, line 53: I find weird to have a comment at the future tense. What do you think about rephrasing it ? Or if this API is not yet ready for being used by publishers, maybe it is worth mentioning that it should not be used (maybe even mark it with @Internal ??)

Also, don't you mind adding a comment on the other constructor explaining that it is intended for standalone ?

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitial.java
```diff
@@ -51,6 +51,13 @@
   @Nullable
   private CriteoInterstitialAdDisplayListener criteoInterstitialAdDisplayListener;
 
+  /**
+   * Will be used for server side bidding and in-house auction
+   */
+  public CriteoInterstitial(@SuppressWarnings("unused") @NonNull Context context) {
```
TurpIF, line 57: Same comment than for the banner constructor

TurpIF (changes requested):

publisher-sdk/src/androidTest/java/com/criteo/publisher/CriteoBannerViewIntegrationTest.java
Amokrane: Sure, though it's not related to my commit :)

TurpIF: Didn't you add the constructor with only the context param ?

Amokrane: Yes and it is the one that is unit tested here.

TurpIF: There you have two parameters, the context and the ad unit, isn't it ?

Amokrane: Indeed, I stand corrected. Removed.

TurpIF (approved):

PR #61, opened 2020-08-12
Activate custom rules on Detekt

TurpIF:
The target is to restrain coding style to have a homogeneous code base.
Also, code smell are detected to avoid potential bugs.

JIRA: EE-1214


.detekt/config.yml
```diff
@@ -506,15 +506,15 @@ style:
...
+    values: ['TODO', 'FIXME', 'STOPSHIP']
     allowedPatterns: ''
   ForbiddenImport:
     active: false
     imports: []
     forbiddenPatterns: ''
   ForbiddenMethodCall:
-    active: false
+    active: true
     methods: ['kotlin.io.println', 'kotlin.io.print']
```
Amokrane, line 518: Do we have legit cases where we need to call `kotlin.io.print` in production code?

PR #63, opened 2020-08-14
Only prefetch once every 30 secs

Amokrane:
Problem:
For some publishers, the rate of bid requests is relatively higher on
Android than on other platforms. Our assumption is that those publishers
use services that cause Criteo to be initialized more than usual: for
instance on APIs lower than 26, it is possible to restart a service
automatically after the process has been killed.

Solution:
Two solutions were considered:
1. Initialize Criteo only if `Criteo#init` is called from a foreground
process. This would prevent `Criteo` to be initialized from a service, but
it has the adverse side effect of having `Criteo` instance stuck on
`DummyCriteo` instance (since `Criteo#getInstance()` does not attempt
any initialization).

2. Introduce rate limiting on bid requests. The bid request rate is
relatively high (compared to other platforms for the same publisher),
but completely normal/average from an absolute perspective (15-20
reqs/hour). For this reason, a simple strategy has been put in place:
only rate limit calls to prefetch by forbidding two consecutive calls to
prefetch.

The threshold representing the "consecutive" aspect can be configured
remotely but is set to 30 secs by default.

JIRA: EE-648

### DIFF
--- a/publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/BidManagerFunctionalTest.java
+++ b/publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/BidManagerFunctionalTest.java
@@ -36,6 +36,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.timeout;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -190,6 +191,42 @@ public class BidManagerFunctionalTest {
     waitForIdleState();
 
     assertShouldCallCdbAndPopulateCacheOnlyOnce(mappedAdUnitsChunks.get(0), slot);
+  }
+
+  @Test
+  public void prefetch_GivenRateLimited_ShouldNotCallCdbAndPopulateCache() throws Exception {
+    when(config.getMinTimeBetweenPrefetchesMillis()).thenReturn(1L);
+    CdbResponseSlot slot = givenMockedCdbRespondingSlot();
+
+    CacheAdUnit cacheAdUnit = sampleAdUnit();
+    AdUnit adUnit = givenMockedAdUnitMappingTo(cacheAdUnit);
+
+    BidManager bidManager = createBidManager();
+    bidManager.prefetch(singletonList(adUnit));
+    waitForIdleState();
+
+    bidManager.prefetch(singletonList(adUnit));
+    waitForIdleState();
+
+    assertShouldCallCdbAndPopulateCacheOnlyOnce(singletonList(cacheAdUnit), slot);
+  }
+
+  @Test
+  public void prefetch_GivenNotRateLimited_ShouldCallCdbAndPopulateCacheTwice() throws Exception {
+    when(config.getMinTimeBetweenPrefetchesMillis()).thenReturn(0L);
+    CdbResponseSlot slot = givenMockedCdbRespondingSlot();
+
+    CacheAdUnit cacheAdUnit = sampleAdUnit();
+    AdUnit adUnit = givenMockedAdUnitMappingTo(cacheAdUnit);
+
+    BidManager bidManager = createBidManager();
+    bidManager.prefetch(singletonList(adUnit));
+    waitForIdleState();
+
+    bidManager.prefetch(singletonList(adUnit));
+    waitForIdleState();
+
+    assertShouldCallCdbAndPopulateCacheTwice(singletonList(cacheAdUnit), slot);
   }
 
   @Test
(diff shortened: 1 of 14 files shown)


publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/network/PubSdkApiIntegrationTest.java
```diff
@@ -385,7 +386,8 @@ private RemoteConfigResponse defaultRemoteConfigResponse() {
         "<html><body style='text-align:center; margin:0px; padding:0px; horizontal-align:center;'><script src=\"%%displayUrl%%\"></script></body></html>",
         "%%adTagData%%",
         "<html><body style='text-align:center; margin:0px; padding:0px; horizontal-align:center;'><script>%%adTagData%%</script></body></html>",
-        true
+        true,
+        null // FIXME: adjust once the remote config is added
```
TurpIF, line 390: If you want , you can already adjust this by changing the handler CdbMock. This will then make the network tests fail until remote config is added.

publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/BidManagerFunctionalTest.java
```diff
@@ -192,6 +193,42 @@ public void prefetch_GivenAdUnits_ShouldCallCdbAndPopulateCache() throws Excepti
...
+    CdbResponseSlot slot = givenMockedCdbRespondingSlot();
+
+    CacheAdUnit cacheAdUnit = sampleAdUnit();
+    AdUnit adUnit = givenMockedAdUnitMappingTo(cacheAdUnit);
+
+    BidManager bidManager = createBidManager();
+    bidManager.prefetch(singletonList(adUnit));
+    waitForIdleState();
+
+    bidManager.prefetch(singletonList(adUnit));
```
TurpIF, line 208: Do you use the fact the `Clock` is mocked to assume that the prefetch is limited ? If yes, what do you think about adding a comment explaining this ?

publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/BidManagerFunctionalTest.java
```diff
@@ -192,6 +193,42 @@ public void prefetch_GivenAdUnits_ShouldCallCdbAndPopulateCache() throws Excepti
     assertShouldCallCdbAndPopulateCacheOnlyOnce(mappedAdUnitsChunks.get(0), slot);
   }
 
+  @Test
+  public void prefetch_GivenRateLimited_ShouldNotCallCdbAndPopulateCache() throws Exception {
```
TurpIF, line 197: Do you mind adding a test showing that after the clock has been advanced, then it is possible to make again a prefetch.

publisher-sdk/src/main/java/com/criteo/publisher/BidManager.java
```diff
@@ -217,12 +225,15 @@ public void onApplicationStopped() {
    * @param adUnits list of ad units to prefetch
    */
   public void prefetch(@NonNull List<AdUnit> adUnits) {
-    List<List<CacheAdUnit>> requestedAdUnitsChunks = adUnitMapper.mapToChunks(adUnits);
+    if (bidPrefetchRateLimiter.canPrefetch()) {
+      bidPrefetchRateLimiter.setPrefetchTime();
```
TurpIF, line 229: Nit: `set` prefix is usually used for setter method and it may be weird to see a `set*` method without argument.
What do you think about renaming it to show that the rate limiter handle this itself ? Like `markPrefetchTime` ?

publisher-sdk/src/main/java/com/criteo/publisher/BidPrefetchRateLimiter.kt
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * This class guards against calling [BidManager.prefetch] many times in a short time span. This protects
+ * against apps with services that are re-spawned automatically (on boot, after process kill) which
+ * could cause too many initializations of [Criteo] in a short amount of time.
+ */
+@OpenForTesting
+@Internal
+class BidPrefetchRateLimiter(
+    private val clock: Clock,
+    private val sharedPreferences: SharedPreferences,
```
TurpIF, line 34: FAR: those two lines look weird 😄

publisher-sdk/src/test/java/com/criteo/publisher/BidPrefetchRateLimiterTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.nhaarman.mockitokotlin2.whenever
+import org.junit.Before
+import org.junit.Test
+import org.mockito.InjectMocks
+import org.mockito.Mock
+import org.mockito.MockitoAnnotations
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class BidPrefetchRateLimiterTest {
```
TurpIF, line 34: Do you think it might be worth to add a test checking for the persistence ? Or are we confident enough on the shared pref ?

publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/integration/ProfileIdFunctionalTest.kt
```diff
@@ -53,8 +56,12 @@ class ProfileIdFunctionalTest {
   @SpyBean
   private lateinit var api: PubSdkApi
 
+  @MockBean
```
TurpIF, line 59: If you use `SpyBean` you could remove all the stubs, except at line 291.
With it, the rate limiter will act normally, and as the shared pref are cleaned between tests, any tests can run a prefetch if it wants to. In case of the line 291, the memory is cleaned, but the persisted data is not in purpose, hence you'll need the stub there.

PR #72, opened 2020-08-18
Merge CriteoInterstitialAdListener and CriteoInterstitialAdDisplayListener

TurpIF:
* [x] Remove the onAdReadyToDisplay listener.
* [x] Remove the onAdFailedToDisplay listener.
* [x] Change onAdReceived to be fired when interstitial bid and then the creative are received
* [x] Change onAdFailedToReceive to be fired when interstitial bid or creative are not available.
* [x] Provide the CriteoInterstitial in the onAdReceived method
* [x] Force Java 8 and use interface + default methods

JIRA: EE-923


CHANGELOG.md
```diff
@@ -11,6 +11,14 @@
 - Breaking changes
   - Remove `Context` parameter from `CriteoInterstitial` constructor
   - Remove `Parcelable` implementation from `AdSize` class
+  - Move `CriteoInterstitialAdDisplayListener#onAdFailedToDisplay(CriteoErrorCode)` to
+   `CriteoInterstitialAdListener#onAdFailedToReceive(CriteoErrorCode)`. This callback is fired
+    when any kind of error happen after requesting an interstitial ad.
```
Amokrane, line 16: This callback is fired when an error happens after requesting an interstitial ad. 

(Any should be used only for uncountable nouns and plurals and when the sentence is a question or a negative).

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@
...
+  - Move `CriteoInterstitialAdDisplayListener#onAdReadyToDisplay()` to
+   `CriteoInterstitialAdListener#onAdReceived(CriteoInterstitial)`. This callback is fired when an
+    interstitial ad is ready to be displayed.
+  - Provide default implementation in interstitial listener. Java 8 is required, see
+  https://developer.android.com/studio/write/java8-support.
```
Amokrane, line 21: Let's make sure to sync with @irwansetiawan to update the documentation.

publisher-sdk/proguard-rules.pro
```diff
@@ -99,10 +99,6 @@
 
 -keep public interface com.criteo.publisher.CriteoBannerAdListener {*;}
 
--keep public interface com.criteo.publisher.CriteoInterstitialAdListener {*;}
```
Amokrane, line 102: Why are you removing `CriteoInterstitialAdListener`? I thought it was the new interface that publishers need to implement?

TurpIF: I'm using `@Keep` instead. Personally, I find it better than proguard rules as:
- it is not tight to any particular techno
- it is really next to the code declaration, so public API are explicit when reading the code.

If you prefer proguard rules, I can revert back this.

CHANGELOG.md
TurpIF: Sure, I'm trying to condensate a quick summary here. But of course we might need to write a doc for helping on the migration.

publisher-sdk/proguard-rules.pro
Amokrane: I am fine with `@Keep` but in this case, we should move everything to `Keep` that way it would be less confusing.

Amokrane (approved):

publisher-sdk/proguard-rules.pro
TurpIF: Done: EE-1248

PR #81, opened 2020-08-25
Fix functional tests after merging interstitial listeners

TurpIF:
The test was expecting two callbacks from the old interstitial
listeners: `onAdReceived` and `onAdReadyToDisplay`. With the merge of
the listeners, there is now only one callback fired when the
interstitial is ready to be displayed.

Note that this was not caught by CI, because during PR, only
deterministic tests are run, and flaky tests, such as this one, are only
run at merge, which is only activated recently on version branches.


TurpIF:
I've run the flaky tests locally via `./gradlew gordon --tests=androidx.test.filters.FlakyTest` and everything is fine now after this PR.

Amokrane (approved):

PR #83, opened 2020-08-25
Bump version to 4.0.0

TurpIF:
The bump is done now, ahead of time to avoid polluting 3.9.0 artifacts
with breaking changes from the v4.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Next version
 
-- Features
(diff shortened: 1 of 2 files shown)


vguerci (changes requested):
Update changelog version as well maybe?

TurpIF:
> Update changelog version as well maybe?

Changelog is already prepared for the v4. The next section is for the 3.10.0 instead (coming when v4 was created from main branch).
I'm tempted to remove the next section here, so when main and v4 will meet again, the merge would have to be done.

vguerci (approved):

PR #89, opened 2020-09-08
Use DummyCriteo if not initialized properly

Amokrane:
Problem:
When an error happens during the initialization path, Criteo instance
will be `null` and any future call to Criteo.getInstance() will throw a
RuntimeException.

Solution:
DummyCriteo is assigned to Criteo singleton object, when any runtime
exception is thrown during initialization. While the integration will
not work, it will prevent the host application from crashing. Logging
will allow us to investigate and correct the root cause with the
publisher.

-----
Guard against null application in CriteoInterstitial

Problem:
An `application` object is required for the SDK to operate properly.
However, nothing prevents CriteoInterstitialActivity from being open even if
`application` is null.

Solution:
CriteoInterstitial APIs must check wehther `application` is null or
return immediately. This is a workaround to prevent apps from crashing
in production but actions need to be taken in the future to:
1. Understand how `application` can be null for publishers.
2. Check whether we can reduce the scope of `application` within the SDk
and use an Activity context when appropriate (in this case, the Activity
context that is passed to CriteoInterstitial could be used).
3. Understand the implications of making the application object
nullable.

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitial.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitial.java
@@ -79,16 +79,23 @@ public class CriteoInterstitial {
   }
 
   public void setCriteoInterstitialAdListener(
-      @Nullable CriteoInterstitialAdListener criteoInterstitialAdListener) {
+      @Nullable CriteoInterstitialAdListener criteoInterstitialAdListener
+  ) {
     this.criteoInterstitialAdListener = criteoInterstitialAdListener;
   }
 
   public void setCriteoInterstitialAdDisplayListener(
-      @Nullable CriteoInterstitialAdDisplayListener criteoInterstitialAdDisplayListener) {
+      @Nullable CriteoInterstitialAdDisplayListener criteoInterstitialAdDisplayListener
+  ) {
     this.criteoInterstitialAdDisplayListener = criteoInterstitialAdDisplayListener;
   }
 
   public void loadAd() {
+    if (!DependencyProvider.getInstance().isApplicationSet()) {
+      Log.w(TAG, "Calling CriteoInterstitial#loadAd with a null application");
+      return;
+    }
+
     try {
       doLoadAd();
     } catch (Throwable tr) {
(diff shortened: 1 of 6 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitial.java
```diff
@@ -79,16 +79,23 @@ public CriteoInterstitial(
...
   public void setCriteoInterstitialAdDisplayListener(
-      @Nullable CriteoInterstitialAdDisplayListener criteoInterstitialAdDisplayListener) {
+      @Nullable CriteoInterstitialAdDisplayListener criteoInterstitialAdDisplayListener
+  ) {
     this.criteoInterstitialAdDisplayListener = criteoInterstitialAdDisplayListener;
   }
 
   public void loadAd() {
+    if (!DependencyProvider.getInstance().isApplicationSet()) {
+      Log.w(TAG, "Calling CriteoInterstitial#loadAd with a null applicationContext");
```
vguerci, line 95: Maybe the message should be `Calling CriteoInterstitial#loadAd with a null application` to be consistent with other ones?

publisher-sdk/src/main/java/com/criteo/publisher/CriteoNotInitializedException.java
```diff
@@ -20,7 +20,7 @@
 
   public CriteoNotInitializedException(String message) {
     super(message + "\n"
-        + "Did you initialize the Criteo SDK ?\n"
+        + "Did you properly initialize the Criteo SDK ?\n"
```
vguerci, line 23: NIT: Why asking a question as if we weren't sure? I would suggest an informative/neutral message instead.

publisher-sdk/src/test/java/com/criteo/publisher/CriteoInterstitialTest.java
```diff
@@ -161,6 +162,61 @@ public void displayAd_GivenController_DelegateToIt() throws Exception {
     verify(controller).fetchCreativeAsync("fake_display_data");
   }
 
+  @Test
+  public void loadAd_GivenNullApplicationContext_ReturnImmediately() throws Exception {
```
vguerci, line 166: NIT: Naming consistency, `GivenNullApplicationContext` vs `givenNullApplication()`

vguerci (changes requested):
Just curious, how this affects only interstitials?

publisher-sdk/src/main/java/com/criteo/publisher/CriteoNotInitializedException.java
Amokrane: Because we are not really sure :) For instance, one of the publishers that reporter a problem in this area did initialize it properly.

vguerci:
Thanks for the explanations

vguerci (approved):

PR #92, opened 2020-09-15
Prevent interstitial activity to crash when started by system with the SDK no initialized

TurpIF:


publisher-sdk/src/main/java/com/criteo/publisher/csm/FileMetricRepository.java
```diff
@@ -104,22 +106,19 @@ boolean contains(@NonNull String impressionId) {
...
   @NonNull
   private SyncMetricFile getOrCreateMetricFile(@NonNull File metricFile) {
-    SyncMetricFile oldMetric = metricFileById.get(metricFile);
-    if (oldMetric == null) {
-      SyncMetricFile newMetric = directory.createSyncMetricFile(metricFile);
-      oldMetric = (oldMetric = metricFileById.putIfAbsent(metricFile, newMetric)) == null
-          ? newMetric : oldMetric;
-    }
-    return oldMetric;
+    return MapUtilKt.getOrCompute(metricFileById, metricFile, new Function0<SyncMetricFile>() {
```
Amokrane, line 116: I would have preferred if we didn't include more changes on this commit than what is necessary. Can you separate the fix for the 3.10.2 and ship all the other changes later?

TurpIF: Ok, I've put this in another commit, so the 3.10.2 could be made only with the two first commits:
- try catch in activity
- thread safe dependency provider

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitialActivity.java
```diff
@@ -36,26 +36,39 @@
...
   private ComponentName callingActivityName;
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
+    try {
+      super.onCreate(savedInstanceState);
+      doOnCreate();
+    } catch (Throwable t) {
+      logger.error("Error while creating interstitial activity.", t);
```
Amokrane, line 58: I find it a bit risky to use the `LoggerFactory` here, because if we ever decide to inject the context when building a `LoggerFactory` then it would crash again. Could you add a comment on the `LoggerFactory` to explain that injecting a context could cause a side effect on `CriteoInterstitialActivity`?

Amokrane (changes requested):
minor request.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInterstitialActivity.java
TurpIF: The test is checking this integration already and if someone make the logger factory crash, then it this test would be red.

Knowing this, do you confirm that you still want to add a comment ?

Amokrane: In this case we are good! :)

Amokrane (approved):

PR #97, opened 2020-09-22
Use time budget from CaC rather than hard coded one

TurpIF:
JIRA: EE-1223

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/model/Config.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/model/Config.java
@@ -58,6 +58,7 @@ public class Config {
     private static final String AD_TAG_DATA_MODE = "<html><body style='text-align:center; margin:0px; padding:0px; horizontal-align:center;'><script>%%adTagData%%</script></body></html>";
     private static final boolean CSM_ENABLED = true;
     private static final boolean LIVE_BIDDING_ENABLED = false;
+    private static final int LIVE_BIDDING_TIME_BUDGET_IN_MILLIS = 5_000;
 
   }
 
@@ -158,6 +159,10 @@ public class Config {
         getOrElse(
             overrideRemoteConfig.getLiveBiddingEnabled(),
             baseRemoteConfig.getLiveBiddingEnabled()
+        ),
+        getOrElse(
+            overrideRemoteConfig.getLiveBiddingTimeBudgetInMillis(),
+            baseRemoteConfig.getLiveBiddingTimeBudgetInMillis()
         )
     );
   }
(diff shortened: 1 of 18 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/model/Config.java
```diff
@@ -58,6 +59,7 @@
     private static final String AD_TAG_DATA_MODE = "<html><body style='text-align:center; margin:0px; padding:0px; horizontal-align:center;'><script>%%adTagData%%</script></body></html>";
     private static final boolean CSM_ENABLED = true;
     private static final boolean LIVE_BIDDING_ENABLED = false;
+    private static final int LIVE_BIDDING_TIME_BUDGET = BuildConfig.networkTimeoutInMillis;
```
Amokrane, line 62: I wouldn't put a default equal to the networktimeout as it's quite big. Instead, I would put 5_000ms which is the baseline that we intend to set.

Amokrane (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/model/RemoteConfigResponse.java
```diff
@@ -149,12 +153,20 @@ public RemoteConfigResponse withKillSwitch(@Nullable Boolean killSwitch) {
...
   @Nullable
   public abstract Boolean getLiveBiddingEnabled();
+
+  /**
+   * Amount of time (in milliseconds) given to the SDK to serve a bid to the publisher. If the SDK
+   * get a CDB response within this time budget, SDK returns it directly. Else, cached bid is used
+   * (if present) and CDB response is cached for later.
+   */
+  @Nullable
+  public abstract Integer getLiveBiddingTimeBudget();
```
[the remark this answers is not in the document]
TurpIF: Ok but I would then need to add a @SerializedName annotation: this method is used for deserializing the CDB response

publisher-sdk/src/main/java/com/criteo/publisher/model/Config.java
TurpIF: Ok

publisher-sdk/src/main/java/com/criteo/publisher/model/RemoteConfigResponse.java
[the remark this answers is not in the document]
Amokrane: I would argue that even the CDB payload should have had this name (sorry if I didn't catch this earlier).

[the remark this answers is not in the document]
TurpIF: Done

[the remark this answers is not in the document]
TurpIF: I will update CDB in parallel

publisher-sdk/src/main/java/com/criteo/publisher/network/LiveBidRequestSender.kt
```diff
@@ -29,19 +30,17 @@ class LiveBidRequestSender(
...
-  }
 
   fun sendLiveBidRequest(
       cacheAdUnit: CacheAdUnit,
       liveCdbCallListener: LiveCdbCallListener
   ) {
     scheduledExecutorService.schedule({
       liveCdbCallListener.onTimeBudgetExceeded()
-    }, DEFAULT_TIME_BUDGET_IN_MILLIS, TimeUnit.MILLISECONDS)
+    }, config.liveBiddingTimeBudgetInMillis.toLong(), TimeUnit.MILLISECONDS)
```
Amokrane, line 43: can you define the default as a long value to prevent this conversion?

Amokrane (changes requested):
minor request

publisher-sdk/src/main/java/com/criteo/publisher/network/LiveBidRequestSender.kt
TurpIF: I'm afraid I am not understanding you. This is not a default value, this is the value given by CDB.

Using an `int` instead of `long` from CaC to `Config#getLiveBiddingTimeBudgetInMilis` is done in purpose: there is no way that such value could be over 2^32.
IMHO we should not chose a technical type just because of one techno forcing us this cast: ie if `ScheduledExecutorService` accepted an `int` then we wouldn't even think about using a `long`. Moreover going from `int` to `long` is lossless.

Amokrane: Ah I read `buildConfig.liveBiddingTimeBudgetInMillis`, nvm.

Amokrane (approved):

PR #98, opened 2020-09-24
Add support for advanced native when live-bidding

TurpIF:
JIRA: EE-1275


publisher-sdk/src/main/java/com/criteo/publisher/BidListener.kt
```diff
@@ -16,14 +16,15 @@
...
+import androidx.annotation.NonNull
 import com.criteo.publisher.model.CdbResponseSlot
 
 /**
  * Callback through which bid responses will be served asynchronously.
  * Each of the callback methods will only be called once.
  */
 interface BidListener {
-  fun onBidResponse(cdbResponseSlot: CdbResponseSlot)
+  fun onBidResponse(@NonNull cdbResponseSlot: CdbResponseSlot)
```
Amokrane, line 27: What's the value of `@NonNull` annotation in Kotlin code?

Amokrane (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/Standalone.kt
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package com.criteo.publisher
+
+import com.criteo.publisher.annotation.OpenForTesting
+import com.criteo.publisher.model.AdUnit
+import com.criteo.publisher.model.Config
+
+@OpenForTesting
+internal class Standalone(
```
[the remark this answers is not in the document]
TurpIF: What about putting this logic in the `BidManager` ? You're right that that this API might be used by any integration but I think that `LiveBidder` is not the right name: it is not always doing live bids.
Moreover, if later we drop the old bidding way, then this method would just delegate to the `BidManager`.

publisher-sdk/src/main/java/com/criteo/publisher/BidListener.kt
TurpIF: image

You're right that's a mistake. My implementation automatically added a `@NonNull` and I was confused why the interface didn't had one. And I didn't get that it was Kotlin code, not Java.

publisher-sdk/src/main/java/com/criteo/publisher/Standalone.kt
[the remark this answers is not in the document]
TurpIF: Done

publisher-sdk/src/main/java/com/criteo/publisher/BidManager.java
```diff
@@ -125,6 +151,7 @@
    * @return a valid bid that may be displayed or <code>null</code> that should be ignored
    */
   @Nullable
+  // @VisibleForTesting TODO Migrate callers on #getBidForAdUnit
```
Amokrane, line 154: Can you elaborate on the TODO more? JIRA ticket?

Amokrane (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/BidManager.java
TurpIF: Sure, updated

Amokrane (approved):

PR #103, opened 2020-09-25
Support live-bidding in AppBidding integration mode

Amokrane:


publisher-sdk/src/main/java/com/criteo/publisher/BidListener.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.annotation.Keep
+
+/**
+ * Interface to be used in `App Bidding` integration mode to know when bidding has completed,
+ * before the Ad can be loaded.
+ *
+ * @see: https://publisherdocs.criteotilt.com/app/android/app-bidding/
+ */
+@Keep
+interface BidListener {
```
TurpIF, line 28: `BidListener` is very generic and could also fit for InHouse. Do you think it might be good to insert the name of the integration to tight the listener ?
Like `AppBiddingListener` and `InHouseListener` ?

This is just a proposition, do not hesitate to confirm that you really want a `BidListener` for AppBidding.

Amokrane: InHouse will actually have a different API. For InHouse it is necessary to return a `BidResponse` object in the callback param. So I was thinking: `BidListener` or `BidCompleteListener` for AppBidding and `BidResponseListener` for InHouse. WDYT?

Amokrane: For InHouse it is necessary to return a BidResponse object in the callback param. So I was thinking: BidListener or `BidCompleteListener` for AppBidding and `BidResponseListener` for InHouse. I think the name of the listener should follow its usage, not the integration mode. WDYT?

TurpIF: Ok then. Those names are precise enough to not lock us with a generic name.

Amokrane: I will update the design doc with the API to get buy-in from all stakeholders.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInternal.java
```diff
@@ -121,16 +121,16 @@ public void runSafely() {
   }
 
   @Override
-  public void setBidsForAdUnit(Object object, AdUnit adUnit) {
+  public void setBidsForAdUnit(Object object, AdUnit adUnit, @NonNull BidListener bidListener) {
     try {
-      doSetBidsForAdUnit(object, adUnit);
+      doSetBidsForAdUnit(object, adUnit, bidListener);
     } catch (Throwable e) {
       Log.e(TAG, "Internal error while setting bids for adUnit.", e);
```
TurpIF, line 128: In the worst case, we can consider that not calling the listener === blocking the publisher

Then it is hard to chose what to do here:
If we have an error, then listener could be not invoked at all. A solution is to call the listener in the catch
But calling the listener in the catch may throw another exception => solution: need to catch it again and use `adSuppressed`.
But with async, calling the listener in the catch may end to call the listener twice => solution: wrap listener in an atomic single shot listener.

Just proposing some solution to issues that may not happen. We can simply wait for reaction of publishers.

TurpIF (changes requested):
Please remove the publisher-sdk/jacoco.exec

publisher-sdk/src/test/java/com/criteo/publisher/headerbidding/HeaderBiddingTest.kt
```diff
@@ -64,7 +66,7 @@ class HeaderBiddingTest {
 
     val headerBidding = HeaderBidding(bidManager, listOf(handler), integrationRegistry)
 
-    headerBidding.enrichBid(obj, adUnit)
+    headerBidding.enrichBid(obj, adUnit, mock())
```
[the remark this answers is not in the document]
Amokrane: That's not what happens currently though? Currently, if there is a nofill, then the `setBidForAdUnit` returns and doesn't do anything, which makes sense because in this case, we don't participate to the server side auction.

I can add a second callback method to the `BidListener` called: onNoBid(), but we will need to think about how it will be used by the publishers. Let's talk about that tomorrow.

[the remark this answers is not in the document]
Amokrane: Yep indeed. Will correct.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInternal.java
Amokrane: Errors (such as timeout) should trigger the listener as well. The listener will be called on the UI thread. Why would it result in calling the listener twice?

TurpIF: By error, I meant `Throwable`. Once could imagine that async task is started and a throwable is thrown then.

Your sample of code for the integration shows that publisher are using a `CountDownLatch` that they will count down everytime the listener is notified.

Here is your sample:
```
private final CountDownLatch latch = new CountDownLatch(TOTAL_COMPETING_SDKS);
criteo.loadAd(publisherAdView, BANNER, new BidCompleteListener() {
     @Override
     public void onBiddingComplete() {
        latch.countDown();
     }
});
competitor.loadAd(publisherAdView, BANNER, new BidCompleteListener() {
     @Override
     public void onBiddingComplete() {
         latch.countDown();
     }
});
 
 
executor.execute(() -> {
    // give a chance to all SDKs to respond, but timeout at 5 secs
    latch.await(5, TimeUnit.SECONDS);
    Handler handler = new Handler(Looper.getMainLooper());
        handler.post(() -> {
          publisherAdView.loadAd();
     });
});
```

Calling twice the listener would make the SDK acts twice.

Still again, it is just a proposition of solution on issues that may be not worth it to handle.

Amokrane: The second call (`competitor.loadAd()`) is from a different Ad SDK and does not call our listener. This example was just meant to show how a publisher could block until all Ad SDKs have finished bidding. As far as our listener is concerned, it will only be called once.

PR #104, opened 2020-09-25
Do not store technical metrics unless TCF consent is given

TurpIF:
JIRA: EE-1261

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/privacy/UserPrivacyUtil.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/privacy/UserPrivacyUtil.java
@@ -148,4 +148,11 @@ public class UserPrivacyUtil {
   public String getMopubConsent() {
     return safeSharedPreferences.getString(MOPUB_CONSENT_SHARED_PREFS_KEY, "");
   }
+
+  /**
+   * Indicate if user didn't give its consent for CSM when applicable.
+   */
+  public boolean isCsmDisallowed() {
+    return false; // TODO
+  }
 }
--- a/publisher-sdk/src/main/java/com/criteo/publisher/privacy/UserPrivacyUtil.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/privacy/UserPrivacyUtil.java
@@ -16,10 +16,8 @@
 
 package com.criteo.publisher.privacy;
 
-import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
-import android.preference.PreferenceManager;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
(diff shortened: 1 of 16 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/privacy/UserPrivacyUtil.java
```diff
@@ -148,4 +152,42 @@ public void storeMopubConsent(@Nullable String mopubConsent) {
   public String getMopubConsent() {
     return safeSharedPreferences.getString(MOPUB_CONSENT_SHARED_PREFS_KEY, "");
   }
+
+  /**
+   * Indicate if user didn't give its consent for CSM when applicable.
+   *
+   * <ul>
+   *   <li>
+   *     IABTCF_PublisherRestrictions{ID}: publisher restrictions on purpose {ID} (only for ID=1)
```
Amokrane, line 161: I feel we should explain somewhere what purpose 1 means. That way, the reader/maintainer understands why we only care about `ID=1`. A link to the list of Purpose IDs would be great.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/UserPrivacyUtil.java
```diff
@@ -148,4 +152,42 @@ public void storeMopubConsent(@Nullable String mopubConsent) {
...
+   * </ul>
+   */
+  public boolean isCsmDisallowed() {
+    TcfGdprStrategy tcfGdprStrategy = tcfStrategyResolver.resolveTcfStrategy();
+    if (tcfGdprStrategy == null) {
+      return false;
+    }
+
+    PublisherRestrictionType restrictionTypeForPurpose1 = tcfGdprStrategy.getPublisherRestrictionTypeForPurpose1();
+    if (restrictionTypeForPurpose1 == NOT_ALLOWED || restrictionTypeForPurpose1 == REQUIRE_LEGITIMATE_INTEREST) {
```
Amokrane, line 185: If legitimate interest is required, shouldn't we check if legitimate interest  is established for `Criteo`? (`IABTCF_VendorLegitimateInterests`)

Amokrane (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/privacy/gdpr/TcfGdprStrategy.java
```diff
@@ -17,16 +17,67 @@
...
 
+  /**
+   * Gives the status of vendor consent for Criteo
+   * <p>
+   * If it is unknown (such as for TCF1), <code>null</code> is returned
+   *
+   * @return <code>true</code> if consent is explicitly given for Criteo vendor
+   */
+  @Nullable
+  default Boolean isVendorConsentGiven() {
```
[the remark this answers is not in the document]
TurpIF: Done, the `TcfGdprStrategy` interface and implementation are too narrow concepts. They purpose is only to provide a `GdprData` for the bidding.
I've added a `Tcf2CsmGuard` dedicated to checking TCF2 for CSM. A `CsmGuard` interface could be extracted and implemented in multiple manner if we get new cases: CCPA, ...

publisher-sdk/src/main/java/com/criteo/publisher/privacy/UserPrivacyUtil.java
TurpIF: From the TCF point of view we're vendor, not publishers. Hence there is no link between the two.
For the publisher restriction, it would be like: "do you consent that a publisher store a personal data". Vs for the vendor it would be: "do you consent that a publisher is sending data to a vendor for ...".

I guess it's complicated legal requirements, and to be honest, I prefer to trust product team on this 😄

Amokrane (dismissed):

Amokrane (approved):

PR #107, opened 2020-09-28
Support dependency provider with lambda capturing "this"

TurpIF:
As we're moving forward on Java 8 feature, lambdas will appear in the
dependency provider. Currently the DependenciesAnnotationInjection does
not handled method generated by Android compiler when a lambda is
capturing `this`.

This is fixed by ignoring synthetic methods. Synthetic elements are the
elements added by a Java compiler but without any trace in the source
code. Hence it is legitimate to ignore such methods.

### DIFF
--- a/test-utils/build.gradle.kts
+++ b/test-utils/build.gradle.kts
@@ -26,9 +26,21 @@ plugins {
 
 androidLibModule()
 
+android {
+  defaultConfig {
+    multiDexEnabled = true
+  }
+
+  packagingOptions {
+    // Both AssertJ and ByteBuddy (via Mockito) bring this and the duplication yields an error
+    exclude("META-INF/licenses/ASM")
+  }
+}
+
 dependencies {
   compileOnly(project(":publisher-sdk"))
 
+  implementation(Deps.AndroidX.MultiDex)
   implementation(Deps.JUnit.JUnit)
   implementation(Deps.Square.OkHttp.MockWebServer)
   compileOnly(Deps.AutoValue.GsonRuntime)
@@ -50,6 +62,8 @@ dependencies {
 
   androidTestImplementation(Deps.AndroidX.Test.Runner)
   androidTestImplementation(Deps.AssertJ.AssertJ)
+  androidTestImplementation(Deps.Mockito.Android)
+  androidTestImplementation(Deps.Mockito.Kotlin)
 
   detektPlugins(Deps.Detekt.DetektFormatting)
 }
--- a/test-utils/src/androidTest/java/com/criteo/publisher/mock/DependenciesAnnotationInjectionAndroidTest.kt
+++ b/test-utils/src/androidTest/java/com/criteo/publisher/mock/DependenciesAnnotationInjectionAndroidTest.kt
@@ -1,0 +1,50 @@
+/*
+ *    Copyright 2020 Criteo
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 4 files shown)


test-utils/build.gradle.kts
```diff
@@ -26,9 +26,21 @@ plugins {
 
 androidLibModule()
 
+android {
+  defaultConfig {
+    multiDexEnabled = true
+  }
+
+  packagingOptions {
+    // Both AssertJ and ByteBuddy (via Mockito) brings this and the duplication yield an error
```
Amokrane, line 35: bring
yields

test-utils/src/androidTest/java/com/criteo/publisher/mock/DependenciesAnnotationInjectionAndroidTest.kt
```diff
@@ -0,0 +1,47 @@
...
+   * When a lambda is capturing only `this`, then a new method is created:
+   * - method is synthetic
+   * - method has no argument
+   * - method returns what the lambda returns
+   *
+   * For [javax.inject.Inject] objects, the [DependenciesAnnotationInjection] is looking for methods without argument
+   * and returning the expected type. So the generated lambda method is a matching candidate.
+   * Unless the injection check candidate methods are not synthetic.
+   *
+   * Note that this is only observable on Android VM. JVM are handling lambdas this correctly.
```
Amokrane, line 34: Android Runtime 

JVMs handle lambdas, that capture `this`, correctly.

test-utils/src/main/java/com/criteo/publisher/mock/DependenciesAnnotationInjection.java
```diff
@@ -289,6 +289,23 @@ private Method findProviderMethod(
...
+         * When a lambda is capturing only `this`, then a new method is created:
+         * - method is synthetic
+         * - method has no argument
+         * - method returns what the lambda returns
+         *
+         * Such generated method may collide with real candidates and should not be considered. So they are ignored by
+         * using the fact that they are synthetic.
+         *
+         * Note that this is only observable on Android VM. JVM are handling lambdas this correctly.
+         *
```
Amokrane, line 303: Android Runtime

JVMs handle lambdas, that capture this, correctly.

test-utils/src/main/java/com/criteo/publisher/mock/DependenciesAnnotationInjection.java
```diff
@@ -289,6 +289,23 @@ private Method findProviderMethod(
...
+        /*
+         * When a lambda is capturing only `this`, then a new method is created:
+         * - method is synthetic
+         * - method has no argument
+         * - method returns what the lambda returns
+         *
+         * Such generated method may collide with real candidates and should not be considered. So they are ignored by
+         * using the fact that they are synthetic.
+         *
+         * Note that this is only observable on Android VM. JVM are handling lambdas this correctly.
```
Amokrane, line 302: what does `correctly` mean here? Does it mean not generating synthetic methods?

Amokrane (changes requested):

test-utils/src/main/java/com/criteo/publisher/mock/DependenciesAnnotationInjection.java
TurpIF: In fact this happen during both compilation/runtime phase. I added more detail on what is done by the Android compiler vs what is done by the Javac+JVM.

Java is just putting an invoke dynamic at call site which will use the [lambda meta factory](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/LambdaMetafactory.html) to generate lazily a SAM and invoke it.
OTOH, Android seems to handle lambda at compilation and generate a method instead.

Amokrane (approved):

PR #109, opened 2020-10-01
Remove BidToken and use BidResponse in public API

TurpIF:
This is a preparation PR to migrate InHouse to live bidding.

The `BidToken` class is removed and only the `BidResponse` object is used in InHouse public API. Moreover, data is inlined in the `BidResponse` so that all classes related to `BidToken` can be removed: `TokenCache`, `TokenValue`, ...

This make the `BidResponse` object looks like the target `Bid` object that would be share between InHouse and AppBidding.

JIRA: EE-1225


vguerci:
`+300 -950` 👀

publisher-sdk/src/main/java/com/criteo/publisher/BidResponse.java
```diff
@@ -17,73 +17,46 @@
 package com.criteo.publisher;
 
 import androidx.annotation.Keep;
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import com.criteo.publisher.model.AbstractTokenValue;
```
Amokrane: FAR: what do you think about renaming `AbstractTokenValue` to `TokenValue`?

publisher-sdk/src/main/java/com/criteo/publisher/BidResponse.java
```diff
@@ -55,8 +64,29 @@ public boolean isBidSuccess() {
...
+  }
+
+  @Nullable
+  private synchronized <T> T consume(Function1<CdbResponseSlot, T> action) {
+    if (slot == null || slot.isExpired(clock)) {
+      return null;
+    }
+
+    T element = action.invoke(slot);
+    slot = null;
```
Amokrane, line 77: I think we need some code-documentatin to explain why we are nulling out the slot.

publisher-sdk/src/main/java/com/criteo/publisher/BidResponse.java
```diff
@@ -19,30 +19,39 @@
...
 
-  protected BidResponse(
+  private final Clock clock;
+
+  @Nullable
+  private CdbResponseSlot slot;
+
+  BidResponse(
       double price,
       boolean valid,
```
Amokrane: Could you refresh my memory on what `valid` is for? Why would a publisher pass `false` here?

Amokrane (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/BidResponse.java
TurpIF: This class is removed in 7ec164d

publisher-sdk/src/main/java/com/criteo/publisher/BidResponse.java
TurpIF: This was used to tell to publisher if the bid response was successful or not. But 609f946 is replacing the `NO_BID` object by `null`. This boolean is then removed in 609f946 too.

publisher-sdk/src/main/java/com/criteo/publisher/BidResponse.java
TurpIF: Sure, done

Amokrane: Is it clear from the publisher point of view that getting a display url consumes the bid? `consumeDisplayUrl` does not indicate that the bid is consumed, but that the displayUrl is consumed. Also, why do we want to have a bid that is usable only once? Is it to workaround the possibility of the publisher caching the bid? But in this case, the publisher can still keep the bid in cache the first time and use it later. I also thought we agreed that it is the backend responsibility to detect when a displayUrl for a given bid is used twice.

@vguerci what are your thoughts on this?

TurpIF: Only the `getPrice` method is visible from publishers. Other are internal methods. Let me kotlinify this class, so I can then use `internal` keywords.

The "usable once" logic was already there on InHouse integration. This PR does not add or remove any functionality. At the base, yes I guess it is a way to mitigate the risk that a publisher could save the bid and reuse it multiple time.
Also, even if we can rely on TLA to handle such duplication, I think it is still worth it to skip trivial errors from publisher at client-side level. This is a way to reduce the infra cost.

TurpIF: I finally used the `@Internal` annotation to explicit that consume* methods are not for publishers.

Amokrane (approved):

PR #110, opened 2020-10-02
Use live bidding with InHouse

TurpIF:
JIRA: EE-1225

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/Bid.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/Bid.java
@@ -27,7 +27,7 @@ import com.criteo.publisher.model.nativeads.NativeAssets;
 import com.criteo.publisher.util.AdUnitType;
 import kotlin.jvm.functions.Function1;
 
-public class BidResponse {
+public class Bid {
 
   private final double price;
 
@@ -40,7 +40,7 @@ public class BidResponse {
   @Nullable
   private CdbResponseSlot slot;
 
-  BidResponse(
+  Bid(
       @NonNull AdUnitType adUnitType,
       @NonNull Clock clock,
       @NonNull CdbResponseSlot slot
--- a/publisher-sdk/src/main/java/com/criteo/publisher/BidResponseListener.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/BidResponseListener.java
@@ -1,0 +1,46 @@
+/*
+ *    Copyright 2020 Criteo
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.criteo.publisher;
+
+import androidx.annotation.Keep;
+import androidx.annotation.Nullable;
+import com.criteo.publisher.advancednative.CriteoNativeLoader;
+
+@Keep
+public interface BidResponseListener {
+
+  /**
+   * Callback invoked when a response for a bid is given to the publisher.
+   * <p>
+   * In case of no bid, the given response is <code>null</code>. In case of successful bid, the given response is not
+   * <code>null</code> and its CPM is accessible through {@link BidResponse#getPrice()}.
+   *
+   * <h1>InHouse integration</h1>
+   * When using the In-House integration, the response can be given to one of these methods, to display the Ad:
+   * <ul>
+   *   <li>{@link CriteoBannerView#loadAd(BidResponse)}: display a banner</li>
+   *   <li>{@link CriteoInterstitial#loadAd(BidResponse)}: display an interstitial</li>
+   *   <li>{@link CriteoNativeLoader#loadAd(BidResponse)}: display a native ad</li>
+   * </ul>
+   * <p>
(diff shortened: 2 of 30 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/BidResponseListener.java
```diff
@@ -0,0 +1,47 @@
...
+   *
+   * <h1>InHouse integration</h1>
+   * When using the In-House integration, the response can be provided to one a those methods to display the ads:
+   * <ul>
+   *   <li>{@link CriteoBannerView#loadAd(BidResponse)}: display a banner</li>
+   *   <li>{@link CriteoInterstitial#loadAd(BidResponse)}: display an interstitial</li>
+   *   <li>{@link CriteoNativeLoader#loadAd(BidResponse)}: display a native ad</li>
+   * </ul>
+   * <p>
+   * Please note that you should use the <code>loadAd</code> method corresponding to the kind of ad unit you asked a bid
```
Amokrane, line 40: Regarding documentation we should be consistent with our tone of voice. Here we are mixing 2nd and 3rd person sentences.

`Callback invoked when a response for a bid is given to the publisher.` -> 3rd person
`Please note that you should use the <code>loadAd</code> method corresponding to the kind of ad unit you asked a bid` -> 2nd person.

Which tone of voice you think is the most appropriate for code documentation?

publisher-sdk/src/main/java/com/criteo/publisher/InHouse.java
```diff
@@ -61,14 +62,39 @@ public void loadBidResponse(
...
     }
 
-    CdbResponseSlot slot = bidManager.getBidForAdUnitAndPrefetch(adUnit);
-    if (slot == null) {
-      bidResponseListener.onResponse(null);
-      return;
+    bidManager.getBidForAdUnit(adUnit, new BidResponseMapper(adUnit, bidResponseListener, clock));
+  }
+
+  private static class BidResponseMapper implements BidListener {
```
Amokrane, line 68: I suggest renaming to `CdbResponseToBidResponseMapper`. Though not s ure why this can't just be a method instead of a class?

Amokrane (commented):
Comments on commit messages:

1st and 2nd commits have the same titles
Do not consider null ad unit in InHouse module -> could you explain why in the commit desc

publisher-sdk/src/main/java/com/criteo/publisher/BidResponseListener.java
```diff
@@ -0,0 +1,47 @@
+/*
+ *    Copyright 2020 Criteo
```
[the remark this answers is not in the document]
TurpIF: Turning this into a Kotlin interface will make it more harder to use from Kotlin. See https://stackoverflow.com/a/58118410/3663982 that explain well the interop Kotlin/Java of interface.
What we're looking for is the functional interface feature of Kotlin 1.4, but I doubt all publishers are on 1.4 yet.

publisher-sdk/src/main/java/com/criteo/publisher/BidResponseListener.java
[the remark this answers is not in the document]
TurpIF: Do in purpose. The `AdUnit` interface is no longer public (kept by proguard). Only the implementation are.

publisher-sdk/src/main/java/com/criteo/publisher/InHouse.java
TurpIF: `BidResponse` will be renamed to `Bid` and will be the object used by both AppBidding and InHouse. Hence the InHouse class will disappear soon (has it has no more logic).

By method, do you mean anonymous class ? If yes I could do that.

publisher-sdk/src/main/java/com/criteo/publisher/BidResponseListener.java
TurpIF: Used 3rd person

publisher-sdk/src/main/java/com/criteo/publisher/Bid.java
```diff
@@ -24,7 +24,7 @@
 import com.criteo.publisher.util.AdUnitType;
 import kotlin.jvm.functions.Function1;
 
-public class BidResponse {
+public class Bid {
```
Amokrane, line 27: Annotate with Keep and remove `BidResponse` from `proguard-rules`.

TurpIF: `Bid#getPrice` is already `@Keep`. So no need to keep `Bid`. As this was done via annotation, there is no change in proguard file.

publisher-sdk/src/main/java/com/criteo/publisher/InHouse.java
Amokrane: Yep seems like an anonymous class can just do the job simply.

TurpIF:
> Comments on commit messages:
> 
> 1st and 2nd commits have the same titles

No, one is migrating the public API while the other is migrating the private one.

> Do not consider null ad unit in InHouse module -> could you explain why in the commit desc

Done

publisher-sdk/src/main/java/com/criteo/publisher/InHouse.java
TurpIF: Done

publisher-sdk/src/main/java/com/criteo/publisher/Bid.java
Amokrane: What about `BidResponse`? We certainly don't need to keep in the proguard file since it doesn't exist anymore. I am talking about these rules:

```
-keep public class com.criteo.publisher.BidResponse
-keepclassmembers class com.criteo.publisher.BidResponse {
   public *;
}
```

TurpIF: This is already removed in f462e55b which is now merged in `main` branch.

Amokrane (approved):

PR #112, opened 2020-10-05
Use live bidding with AppBidding

TurpIF:
JIRa: EE-1224

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/ConsumableBidLoader.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/ConsumableBidLoader.java
@@ -21,7 +21,13 @@ import com.criteo.publisher.model.AdUnit;
 import com.criteo.publisher.model.CdbResponseSlot;
 import org.jetbrains.annotations.NotNull;
 
-public class InHouse {
+/**
+ * Component delivering consumable {@linkplain Bid bids} for publishers.
+ *
+ * The bids are "consumable" because they can be used only once to display an Ad. Note that {@link Bid} is not called
+ * <code>ConsumableBid</code> because it is part of the public API.
+ */
+public class ConsumableBidLoader {
 
   @NonNull
   private final BidManager bidManager;
@@ -29,7 +35,7 @@ public class InHouse {
   @NonNull
   private final Clock clock;
 
-  public InHouse(
+  public ConsumableBidLoader(
       @NonNull BidManager bidManager,
       @NonNull Clock clock
   ) {
--- a/publisher-sdk/src/main/java/com/criteo/publisher/CriteoInternal.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/CriteoInternal.java
@@ -146,7 +146,7 @@ final class CriteoInternal extends Criteo {
       @NonNull BidResponseListener bidResponseListener
   ) {
     try {
-      inHouse.loadBidResponse(adUnit, bidResponseListener);
+      inHouse.loadBid(adUnit, bidResponseListener);
     } catch (Throwable e) {
       Log.e(TAG, "Internal error while loading bid response.", e);
       bidResponseListener.onResponse(null);
--- a/publisher-sdk/src/main/java/com/criteo/publisher/CriteoInternal.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/CriteoInternal.java
@@ -119,7 +119,7 @@ final class CriteoInternal extends Criteo {
   }
 
   @Override
-  public void setBidsForAdUnit(Object object, @Nullable Bid bid) {
+  public void enrichAdObjectWithBid(Object object, @Nullable Bid bid) {
     try {
       doSetBidsForAdUnit(object, bid);
     } catch (Throwable e) {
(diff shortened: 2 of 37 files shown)


Amokrane:
Few nitpicks regarding commit messages. The first one, in particular, can be a bit more concise.

**Commit#1**:
The interstitial format is not available if the
`CriteoInterstitialActivity` activity is ~~only~~ not registered correctly in the
Android manifest.

This occurred ~~before~~ when publishers w~h~ere using distributed AAR files.
Now that the SDK is distributed only through Maven repositories, such
unavailability should not happen.

This is not totally verified and checking this via remote logs could
determine~d~ if it is a useless check.

~Anyway it is expected to almost never happens~ (already mentioned above). This commit might make
Criteo bids for nothing (what does bidding for nothing means?). Although, the purpose of this is to make the
`Criteo#loadBid` API shar**e**able ~~for~~ between InHouse and AppBidding (after Although..., I expect an explanation. Eg: Although I am tired, I will continue fighting to the end). ~~So~~ Therefore, anything
related to InHouse should be moved from that call.

**Commit#2**:
This is to make the `loadBid` method shar**e**able between both InHouse and
AppBidding.


**Commit#4**:
This component ~~has no more to do~~ has nothing to with InHouse integration anymore. Instead, it
is just managing the bid served directly to publishers. The
particularity of such bid objects is that they are consumable and valid
only once.

TurpIF:
> Few nitpicks regarding commit messages. The first one, in particular, can be a bit more concise.
> 
> **Commit#1**:
> 
> ```
> The interstitial format is not available if the
> `CriteoInterstitialActivity` activity is ~~only~~ not registered correctly in the
> Android manifest.
> This occurred ~~before~~ when publishers w~h~ere using distributed AAR files.
> Now that the SDK is distributed only through Maven repositories, such
> unavailability should not happen.
> 
> This is not totally verified and checking this via remote logs could
> determine~d~ if it is a useless check.
> 
> ~Anyway it is expected to almost never happens~ (already mentioned above). This commit might make
> Criteo bids for nothing (what does bidding for nothing means?). Although, the purpose of this is to make the
> `Criteo#loadBid` API shar**e**able ~~for~~ between InHouse and AppBidding (after Although..., I expect an explanation. Eg: Although I am tired, I will continue fighting to the end). ~~So~~ Therefore, anything
> related to InHouse should be moved from that call.
> ```
> 
> The last two sentences can probably be rewritten into something more concise (straight to the point, avoid repetitions of ideas, remove useless words).
> 
> **Commit#2**:
> 
> ```
> This is to make the `loadBid` method shar**e**able between both InHouse and
> AppBidding.
> ```
> 
> **Commit#4**:
> 
> ```
> This component ~~has no more to do~~ has nothing to with InHouse integration anymore. Instead, it
> is just managing the bid served directly to publishers. The
> particularity of such bid objects is that they are consumable and valid
> only once.
> ```

Done

publisher-sdk/src/main/java/com/criteo/publisher/ConsumableBidLoader.java
```diff
@@ -21,15 +21,15 @@
 import com.criteo.publisher.model.CdbResponseSlot;
 import org.jetbrains.annotations.NotNull;
 
-public class InHouse {
+public class ConsumableBidLoader {
```
Amokrane, line 24: I think it's worth documenting why this class has such a name (even if the class is internal). Today we interpret that `consuming` as: being able to use the `Bid` only once. But, if another person comes in this won't be entirely clear and she will think: what does an unconsumable bid mean?

Amokrane (commented):
Are there tests that cover the time-budget logic for `AppBidding` (similar question for `InHouse`)? Or do we rely on the fact that BidManager & whatever is below it is fully tested?

publisher-sdk/src/main/java/com/criteo/publisher/ConsumableBidLoader.java
TurpIF: Done

publisher-sdk/src/main/java/com/criteo/publisher/CriteoInternal.java
```diff
@@ -119,21 +119,20 @@ public void runSafely() {
   }
 
   @Override
-  public void setBidsForAdUnit(Object object, @NonNull AdUnit adUnit) {
+  public void setBidsForAdUnit(Object object, @Nullable Bid bid) {
```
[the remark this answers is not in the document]
TurpIF: Indeed, I almost forgot this.
I added a commit with the renaming.

TurpIF:
> Are there tests that cover the time-budget logic for `AppBidding` (similar question for `InHouse`)? Or do we rely on the fact that BidManager & whatever is below it is fully tested ?

For the moment we strongly rely on the fact that BidManager is fully tested and AppBidding & InHouse are unit tested.
However, I've just ran the test with live bidding enabled, and bid is impossible: we forgot to set the time of download when live bidding is on.
So I would like to add, at least, tests validating that a bid is possible.

As you prefer, I can either add tests right now with the fix of the time of download, or push a new PR with tests+fix

Amokrane:
> As you prefer, I can either add tests right now with the fix of the time of download, or push a new PR with tests+fix

Whatever makes to you is fine by me :)

Amokrane (approved):

TurpIF:
> > As you prefer, I can either add tests right now with the fix of the time of download, or push a new PR with tests+fix
> 
> Whatever makes sense to you is fine by me :) I have a slight preference for a new PR.

I agree, this one is big enough. I will make a new one.

PR #114, opened 2020-10-07
Make CriteoBannerView a custom view class

Amokrane:
In order to use `CriteoBannerView`, publishers have to instantiate it by
hand and add it to a parent layout programmatically. This is both
inefficient as it creates an unnecessary depth in the view tree, and
tedious.

By turning `CriteoBannerView` into a custom view, publishers will be
able to define it in the layout XML file directly.

### DIFF
--- a/app/src/main/java/com/criteo/testapp/InHouseActivity.java
+++ b/app/src/main/java/com/criteo/testapp/InHouseActivity.java
@@ -19,11 +19,10 @@ package com.criteo.testapp;
 import static com.criteo.testapp.PubSdkDemoApplication.INTERSTITIAL_IBV_DEMO;
 import static com.criteo.testapp.PubSdkDemoApplication.NATIVE;
 
-import android.content.Context;
 import android.os.Bundle;
 import android.util.Log;
 import android.widget.Button;
-import android.widget.LinearLayout;
+import android.widget.FrameLayout;
 import androidx.appcompat.app.AppCompatActivity;
 import com.criteo.publisher.Criteo;
 import com.criteo.publisher.CriteoBannerView;
@@ -47,11 +46,11 @@ public class InHouseActivity extends AppCompatActivity {
 
   public static final BannerAdUnit BANNER = new BannerAdUnit(
       "/140800857/Endeavour_320x50",
-      new AdSize(320, 50));
+      new AdSize(320, 50)
+  );
 
-  private Context context;
-  private LinearLayout adLayout;
   private CriteoBannerView criteoBannerView;
+  private FrameLayout nativeAdContainer;
   private CriteoNativeLoader nativeLoader;
   private Button btnShowInterstitial;
   private Button btnShowInterstitialIbv;
@@ -62,14 +61,17 @@ public class InHouseActivity extends AppCompatActivity {
     setContentView(R.layout.activity_in_house);
     MockedIntegrationRegistry.force(Integration.IN_HOUSE);
 
-    adLayout = findViewById(R.id.AdLayout);
+    criteoBannerView = findViewById(R.id.criteoBannerView);
+    nativeAdContainer = findViewById(R.id.nativeAdContainer);
+    criteoBannerView.setCriteoBannerAdListener(new TestAppBannerAdListener(
+        TAG, "In-House"));
+
     btnShowInterstitial = findViewById(R.id.buttonInhouseInterstitial);
     btnShowInterstitialIbv = findViewById(R.id.buttonInhouseInterstitialIbv);
-    context = getApplicationContext();
 
     nativeLoader = new CriteoNativeLoader(
         NATIVE,
-        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), adLayout),
+        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), nativeAdContainer),
         new TestAppNativeRenderer()
     );
 
@@ -101,14 +103,6 @@ public class InHouseActivity extends AppCompatActivity {
   }
 
   private void loadBannerAd() {
-    if (criteoBannerView != null) {
-      criteoBannerView.destroy();
-    }
-
-    criteoBannerView = new CriteoBannerView(context);
-    criteoBannerView.setCriteoBannerAdListener(new TestAppBannerAdListener(
-        TAG, "In-House", adLayout));
-
     Log.d(TAG, "Banner Requested");
     Criteo.getInstance().loadBid(BANNER, criteoBannerView::loadAd);
   }
--- a/app/src/main/java/com/criteo/testapp/StandaloneActivity.java
+++ b/app/src/main/java/com/criteo/testapp/StandaloneActivity.java
@@ -18,13 +18,11 @@ package com.criteo.testapp;
 
 import static com.criteo.testapp.PubSdkDemoApplication.INTERSTITIAL_IBV_DEMO;
 import static com.criteo.testapp.PubSdkDemoApplication.NATIVE;
-import static com.criteo.testapp.PubSdkDemoApplication.STANDALONE_BANNER;
 
-import android.content.Context;
 import android.os.Bundle;
 import android.util.Log;
 import android.widget.Button;
-import android.widget.LinearLayout;
+import android.widget.FrameLayout;
 import androidx.appcompat.app.AppCompatActivity;
 import com.criteo.publisher.CriteoBannerView;
 import com.criteo.publisher.CriteoInterstitial;
@@ -44,12 +42,11 @@ public class StandaloneActivity extends AppCompatActivity {
   private static final InterstitialAdUnit INTERSTITIAL = new InterstitialAdUnit(
       "/140800857/Endeavour_Interstitial_320x480");
 
-  private Context context;
-  private LinearLayout adLayout;
   private CriteoBannerView criteoBannerView;
   private CriteoNativeLoader nativeLoader;
   private Button btnShowInterstitial;
   private Button btnShowInterstitialIbv;
+  private FrameLayout nativeAdContainer;
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
@@ -57,20 +54,27 @@ public class StandaloneActivity extends AppCompatActivity {
     setContentView(R.layout.activity_stand_alone);
     MockedIntegrationRegistry.force(Integration.STANDALONE);
 
-    adLayout = findViewById(R.id.AdLayout);
     btnShowInterstitial = findViewById(R.id.buttonStandAloneInterstitial);
     btnShowInterstitialIbv = findViewById(R.id.buttonStandAloneInterstitialIbv);
+    criteoBannerView = findViewById(R.id.criteoBannerView);
+    nativeAdContainer = findViewById(R.id.nativeAdContainer);
 
-    context = getApplicationContext();
+    criteoBannerView.setCriteoBannerAdListener(new TestAppBannerAdListener(
+        TAG, "Standalone"));
 
     nativeLoader = new CriteoNativeLoader(
         NATIVE,
-        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), adLayout),
+        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), nativeAdContainer),
         new TestAppNativeRenderer()
     );
 
-    findViewById(R.id.buttonStandAloneBanner).setOnClickListener(v -> loadBanner());
+    findViewById(R.id.buttonStandAloneBanner).setOnClickListener(v -> loadBannerAd());
     findViewById(R.id.buttonStandAloneNative).setOnClickListener(v -> loadNative());
+  }
+
+  private void loadBannerAd() {
+    Log.d(TAG, "Banner Requested");
+    criteoBannerView.loadAd();
   }
 
   @Override
@@ -94,19 +98,6 @@ public class StandaloneActivity extends AppCompatActivity {
     if (interstitial.isAdLoaded()) {
       interstitial.show();
     }
-  }
-
-  private void loadBanner() {
-    if (criteoBannerView != null) {
-      criteoBannerView.destroy();
-    }
-
-    criteoBannerView = new CriteoBannerView(context, STANDALONE_BANNER);
-    criteoBannerView.setCriteoBannerAdListener(new TestAppBannerAdListener(
-        TAG, "Standalone", adLayout));
-
-    Log.d(TAG, "Banner Requested");
-    criteoBannerView.loadAd();
   }
 
   private void loadNative() {
--- a/app/src/main/java/com/criteo/testapp/listener/TestAppNativeAdListener.java
+++ b/app/src/main/java/com/criteo/testapp/listener/TestAppNativeAdListener.java
@@ -20,6 +20,7 @@ import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import com.criteo.publisher.CriteoErrorCode;
 import com.criteo.publisher.advancednative.CriteoNativeAd;
 import com.criteo.publisher.advancednative.CriteoNativeAdListener;
@@ -30,7 +31,7 @@ public class TestAppNativeAdListener extends CriteoNativeAdListener {
   private final String prefix;
   private final ViewGroup adLayout;
 
-  public TestAppNativeAdListener(String tag, String prefix, ViewGroup adLayout) {
+  public TestAppNativeAdListener(String tag, String prefix, @Nullable ViewGroup adLayout) {
     this.tag = tag;
     this.prefix = prefix;
     this.adLayout = adLayout;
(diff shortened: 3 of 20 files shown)


app/src/main/java/com/criteo/testapp/listener/TestAppNativeAdListener.java
```diff
@@ -28,21 +28,15 @@
...
     this.tag = tag;
     this.prefix = prefix;
-    this.adLayout = adLayout;
   }
 
   @Override
   public void onAdReceived(@NonNull CriteoNativeAd nativeAd) {
     Log.d(tag, prefix + " - Native onAdReceived");
-
-    View view = nativeAd.createNativeRenderedView(adLayout.getContext(), adLayout);
```
TurpIF, line 43: By removing this, the advanced native is broken. Nothing is shown in the activity.

app/src/main/java/com/criteo/testapp/InHouseActivity.java
```diff
@@ -62,18 +61,25 @@ protected void onCreate(Bundle savedInstanceState) {
...
         NATIVE,
-        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), adLayout),
+        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId()),
         new TestAppNativeRenderer()
     );
 
-    findViewById(R.id.buttonInhouseBanner).setOnClickListener(v -> loadBannerAd());
+    findViewById(R.id.buttonInhouseBanner).setOnClickListener(v -> {
+      Log.d(TAG, "Banner Requested");
+      Criteo.getInstance().loadBid(BANNER, criteoBannerView::loadAd);
```
TurpIF, line 80: You have inlined the `loadBannerAd` but the method is still here. Personally I perfer to have really short lambda to have a more readable code at call site. So I would suggest to revert this line.

app/src/main/java/com/criteo/testapp/StandaloneActivity.java
```diff
@@ -57,19 +54,24 @@ protected void onCreate(Bundle savedInstanceState) {
...
     nativeLoader = new CriteoNativeLoader(
         NATIVE,
-        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), adLayout),
+        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId()),
         new TestAppNativeRenderer()
     );
 
-    findViewById(R.id.buttonStandAloneBanner).setOnClickListener(v -> loadBanner());
+    findViewById(R.id.buttonStandAloneBanner).setOnClickListener(v -> {
+      Log.d(TAG, "Banner Requested");
```
TurpIF, line 72: Nit: Same comment than in `InHouseActivity`. IMHO Having really light lambda keep the call site more readability.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -25,14 +27,15 @@
...
+import com.criteo.publisher.model.AdSize;
 import com.criteo.publisher.model.BannerAdUnit;
 
 public class CriteoBannerView extends WebView {
 
   private static final String TAG = CriteoBannerView.class.getSimpleName();
 
   @Nullable
-  private final BannerAdUnit bannerAdUnit;
+  private BannerAdUnit bannerAdUnit;
```
TurpIF, line 38: IMHO `final` is one of the best keyword in Java. It allows to have the strong guarantee that a value is constant.
You can keep this `final` by setting `bannerAdUnit = null` in the new constructor.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -41,14 +44,39 @@
    * having to init the SDK before.
    */
   @Nullable
-  private final Criteo criteo;
+  private Criteo criteo;
```
TurpIF, line 47: Same than above. You just need to initialize it to keep it `final`

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -41,14 +44,39 @@
...
   @Nullable
   private CriteoBannerAdListener criteoBannerAdListener;
 
   @Nullable
   private CriteoBannerEventController criteoBannerEventController;
 
+  /**
+   * Used when setting {@link CriteoBannerView} in XML
+   */
+  public CriteoBannerView(@NonNull Context context, AttributeSet attrs) {
```
TurpIF, line 58: Shouldn't you need to override the 3 constructors for XML ?
- `constructor(context, attributeSet)`
- `constructor(context, attributeSet, defStyle)`
- `constructor(context, attributeSet, defStyleAttr, defStyleRes)`

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -41,14 +44,39 @@
...
+        0,
+        0
+    );
+
+    try {
+      int width = a.getInteger(R.styleable.CriteoBannerView_criteoAdUnitWidth, -1);
+      int height = a.getInteger(R.styleable.CriteoBannerView_criteoAdUnitHeight, -1);
+      String adUnitId = a.getString(R.styleable.CriteoBannerView_criteoAdUnitId);
+
+      if (adUnitId != null && width >= 0 && height >= 0) {
```
TurpIF, line 72: What do you think about being strict on the input: when at least one field is given, then standalone is targetted and an exception is thrown if: `adUnitId == null || width < 0 || height < 0`. When no field is given, then InHouse is targetted and no check is done.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -25,14 +27,15 @@
 import androidx.annotation.VisibleForTesting;
 import com.criteo.publisher.integration.Integration;
 import com.criteo.publisher.integration.IntegrationRegistry;
+import com.criteo.publisher.model.AdSize;
 import com.criteo.publisher.model.BannerAdUnit;
 
 public class CriteoBannerView extends WebView {
```
TurpIF, line 33: What do you think about adding automated on the inflation from XML ?
You can take a look at `CriteoMediaViewTest` for instance.

TurpIF (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: Actually, I don't remember the purpose of this field? When is it initialized? It seems calling getCriteo (https://github.com/criteo/android-publisher-sdk/blob/main/publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java#L129) will always call `Criteo.getInstance()`, no?

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: Only the first one is needed as there is no need to apply custom styling/theming.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: I think it makes the output uncertain and adds unneeded complexity. For InHouse the field that will be set via `XML` will not even be taken into account because `loadBid` takes an `AdUnit` as a param which will be used instead.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
TurpIF: It is used for unit tests by giving a Criteo mock directly.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
TurpIF: For InHouse there is no issue. But for Standalone, a publisher setting a wrong setting will get its `BannerAdUnit` nullified silently and would only get no bid. Throwing early here would make it clear that input is incorrect.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: Sure, will add a test in another commit on this same PR.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: Yet there is no guarantee that the publisher is using `InHouse` if all three fields are not present.  I would find it confusing to throw only if (!a || !b || !c) and not if (!a && !b && !c). I would rather have a good log that informs the publisher about the initialization status than throw here.

Amokrane: Something like:

```
        logger.warn(
            "The AdUnit associated to this CriteoBannerView instance was not initialized as: "
                + "criteoAdUnitId, criteoAdUnitHeight and criteoAdUnitWidth were not all set. If you are" 
                + "using CriteoBannerView as part of the InHouse integration then this is normal. " 
                + "Otherwise, make sure to set all three attributes."
        );
```

TurpIF: Ok for having a log to indicate the expected integration. However, I still think that the intermediate state is an error.
Isn't it possible to have:
- all correctly set -> `logger.info("CriteoBannerView inflated for Standalone integration.")`
- no setting -> `logger.info("CriteoBannerView inflated for InHouse integration.")`
- missing or incorrect -> throw explaining that InHouse expect nothing and Standalone expect all.

Note that I'm using a log INFO here: InHouse is the second most used integration. We shouldn't warn so often for a normal case.

Amokrane: Alright, that's a good suggestion. Another option would be to have dedicated views per integration: `InHouseCriteoBannerView` and `StandaloneCriteoBannerView`. That would make these widgets more cohesive. Naming could be fine-tuned. WDYT?

TurpIF: Indeed it would make it simpler at implementation level.

OTOH, we would end with `integration * format` different objects. Integration is `{Standalone, InHouse, ServerBidding}` and format is `{Banner, Interstitial, Native}`. That would already make 9 objects almost identical :( For sure it would be possible to use `abstract` classes to factorize the mess but it would still make 9 objects so that's not simpler at SDK level.

I'm not fan of what we have currently. What you're proposing is a lot of work, and I want to be sure that it is worth it.

Amokrane: Ok `lot of work` was my intuition as well. Let's keep it simple then.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
```diff
@@ -49,6 +59,51 @@
...
+      if (adUnitId != null && width != UNSET_DIMENSION_VALUE && height != UNSET_DIMENSION_VALUE) {
+        logger.info("CriteoBannerView inflated for Standalone integration.");
+        bannerAdUnit = new BannerAdUnit(adUnitId, new AdSize(width, height));
+      } else if (adUnitId == null && width == UNSET_DIMENSION_VALUE
+          && height == UNSET_DIMENSION_VALUE) {
+        logger.info("CriteoBannerView inflated for InHouse integration.");
+        bannerAdUnit = null;
+      } else {
+        bannerAdUnit = null;
+        PreconditionsUtil.throwOrLog(new IllegalStateException(
```
TurpIF, line 96: throwOrLog doesn't throw in production. Is this expected ? Why not throwing directly ?

app/src/main/java/com/criteo/testapp/InHouseActivity.java
```diff
@@ -47,10 +46,10 @@
 
   public static final BannerAdUnit BANNER = new BannerAdUnit(
       "/140800857/Endeavour_320x50",
-      new AdSize(320, 50));
+      new AdSize(320, 50)
+  );
 
   private Context context;
```
TurpIF, line 52: Context can be removed. It is not used anymore.

app/src/main/java/com/criteo/testapp/StandaloneActivity.java
```diff
@@ -45,7 +43,6 @@
       "/140800857/Endeavour_Interstitial_320x480");
 
   private Context context;
```
TurpIF, line 45: Context can be removed, it is not used anymore

TurpIF (changes requested):
It is not possible to use both advanced native and banner.

On both Standalone & InHouse activity:
- Open activity
- Click on banner button
- Then banner is displayed correctly

- Open activity
- Click on advanced native button
- Then advanced native is displayed correctly

But

- Open activity
- Click on advanced native button
- Click on banner button
- Expected: banner is displayed
- Actual:advance native is still displayed

TurpIF:
I would also suggest to take a look at the sonar report.

publisher-sdk/src/main/java/com/criteo/publisher/CriteoBannerView.java
Amokrane: We should never throw in production. Throwing unchecked exceptions is only for debugging purposes, apps should never crash in production due to Criteo.

TurpIF: I think we should never throw in production due to variable data coming from publishers or TLA. In this case, the data is constant and the situation is clearly an error. Like when the publisher gives a null cpId, the SDK is crashing during the init to indicate clearly that the integration is wrong.

Amokrane: > Like when the publisher gives a null cpId, the SDK is crashing during the init to indicate clearly that the integration is wrong.

It is no longer the case: https://github.com/criteo/android-publisher-sdk/commit/d3f90a182ca314e6f8e30a8fffef850af1156841#diff-ed8fa8b89a351ac154181bc558d65aa2a98aa2b3e4f861c13ffd332f5277025dR119

> In this case, the data is constant and the situation is clearly an error.
Yes, but Criteo isn't essential for the app to function properly. While publishers will miss some revenue due to this, it is not a reason for crashing the app in production. Instead these cases should be caught with logging.

Making an app crash is really extreme. Imagine the impact from an end user perspective: you are on a shopping application and you add an item to the basket.  Then, when going on the basket page, the app crashes because the banner isn't setup properly. How would you react as a user? (I would rate the app 1 star). How would the publisher reacts if it sees that Criteo is one of their top crashes? (as a publisher I would convince the sales team to drop the SDK showing the bad impact it had on the app's rating and buying rates). So crashing in debug mode -> Yes that's good, it enables the publishers to fix an issue before going to prod. Crashing in production? Must not happen under any circumstances in our case. I would of course throw IllegalStateException if the app wasn't in a proper state to continue running properly, but showing an Ad doesn't count as `important to function properly`.

TurpIF: I expected publishers to start their app and see if they crash.
But ok to not throw then.

TurpIF: Publishers don't have access to the debug mode.
Well, they have if they use `com.criteo.publisher:criteo-publisher-sdk-debug`. But I doubt that anyone is using it.

app/src/main/java/com/criteo/testapp/InHouseActivity.java
```diff
@@ -62,14 +59,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        TAG, "In-House"));
+
     btnShowInterstitial = findViewById(R.id.buttonInhouseInterstitial);
     btnShowInterstitialIbv = findViewById(R.id.buttonInhouseInterstitialIbv);
-    context = getApplicationContext();
 
     nativeLoader = new CriteoNativeLoader(
         NATIVE,
-        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), adLayout),
+        new TestAppNativeAdListener(TAG, NATIVE.getAdUnitId(), criteoBannerView),
```
TurpIF, line 71: Shouldn't you add a native ad layout like for standalone ?
Currently it is not possible to show banner after a native.

app/src/main/res/layout/activity_stand_alone.xml
```diff
@@ -60,15 +60,22 @@
...
           android:background="@android:color/holo_blue_dark"
-          android:text="@string/btn_advanced_native"/>
+          android:text="@string/btn_advanced_native" />
       </LinearLayout>
     </HorizontalScrollView>
-    <LinearLayout
-      android:id="@+id/AdLayout"
+    <com.criteo.publisher.CriteoBannerView
+      android:id="@+id/criteoBannerView"
       android:layout_width="match_parent"
```
TurpIF, line 68: What do you think about using a fixed size ? So it is possible to have both banner and native without changing the visibility.

TurpIF (changes requested):

app/src/main/java/com/criteo/testapp/InHouseActivity.java
Amokrane: I must have sent the wrong patch set. I am passing the `adContainer`. Let me push again.

TurpIF (approved):

PR #116, opened 2020-10-08
[LiveBidding] Add functional tests on BidManager

TurpIF:
JIRA: EE-1224


Amokrane:
Nitpick regarding commit messages. Pre-prending commit titles with [XXX][XXX] breaks the convention that we have for the repo. Commit titles should start with an imperative verb.
Example:
`[BidManager][LiveBidding] Functional tests for valid bids` -> `Test valid bids scenarios when live bidding`.

Amokrane (approved):

Issue #120, opened 2020-10-14
get version number


MDSergioMD:
Hi, is there any way to get the version number at runtime ?

TurpIF:
Hi,

For now it is not possible to get it at runtime via the SDK, but we're planning it.

Meanwhile you can inject it in your `BuidConfig` class with your `build.gradle` file.
Please see this example: https://github.com/criteo/android-publisher-sdk-mopub-adapters/blob/master/mediation/build.gradle#L51. It is injecting a `BuildConfig.SDK_VERSION` constant accessible from the code.

MDSergioMD:
Thank you very much for the reply and help

TurpIF:
Also, I'm wondering why do you need the version during the runtime ?

MDSergioMD:
Hi, we use it to report the version number of our ad providers to our
servers.

Amokrane:
Thanks for the information @MDSergioMD. You will be able to get the version number at runtime with the next version (4.0).

MDSergioMD:
Great news, thank you

PR #123, opened 2020-10-16
#120 Provide version of the SDK at runtime

TurpIF:


Amokrane:
LGTM, but would be good to know for what purpose we are adding such API.

Amokrane (dismissed):

Amokrane (approved):

PR #129, opened 2020-10-22
Set parentActivityName on all test app Activities

Amokrane:


TurpIF (approved):
THANK YOU !!!!!!!!!!!

PR #132, opened 2020-10-22
Fix memory leak in test app

Amokrane:
Memory leak caused by the following reference chain:
CriteoResultReceiver -> .. ->  TestAppInterstitialAdListener -> AppCompatButton  -> StandaloneActivity


TurpIF (approved):
LGTM technically

Although, I dislike the logic on the interstitial button: loading the bid when activity is opened and showing when button is clicked.
I guess that this was done because there was a distinction between when bid is received and when ad is received.

Now that this distinction is erased (listener is notified for success only when bid is received AND ad is received), what do you think about simplifying the logic: one click => ask for bid and if succeed => interstitial is shown. Like for banner or native.

This way, the listener won't have to keep any reference to any view as button is always enabled.

Amokrane:
@TurpIF  Sure, feel free to open a JIRA ticket  (this PR is really just to solve the memory leak issue).

PR #137, opened 2020-10-28
Make connectedCheck work again

Amokrane:
When running `./gradlew connectedCheck`, the following error was
triggered:

More than one file was found with OS independent path 'META-INF/mediation_release.kotlin_module'

Excluding this file is safe.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -36,6 +36,10 @@ android {
         }
     }
 
+    packagingOptions {
+        exclude("META-INF/mediation_release.kotlin_module")
+    }
+
     defaultConfig {
         multiDexEnabled = true
     }


Amokrane:
Not sure why it works on CI but not on my machine, but I consistently get the aforementioned error. With this change it works fine all the time.

TurpIF (commented):
If this is working, maybe could you replace all the references of `:publisher-sdk:connectedCheck` to `connectedCheck`.
The CI is only testing the publisher-sdk module.

Amokrane:
@TurpIF  Interesting. Though in the current workflow files, I see `./gradlew connectedCheck` and not `./gradlew :publisher-sdk:connectedCheck`. I don't see any references of `./gradlew :publisher-sdk:connectedCheck` in the code.

TurpIF:
> @TurpIF Interesting. Though in the current workflow files, I see `./gradlew connectedCheck` and not `./gradlew :publisher-sdk:connectedCheck`. I don't see any references of `./gradlew :publisher-sdk:connectedCheck` in the code.

Ah, I forgot that I already fixed that: https://github.com/criteo/android-publisher-sdk/commit/07df1b7b64edf514b0672d4fe77c14a1b34493ae

TurpIF (approved):

PR #141, opened 2020-10-29
Don't merge: Investigate sonarqube issue

Amokrane:


.github/workflows/main.yml
```diff
@@ -166,7 +166,7 @@ jobs:
           name: android-quality-reports
 
       - name: SonarQube Analysis
-        run: ./gradlew sonarqube
+        run: ./gradlew clean sonarqube --stacktrace
```
TurpIF, line 169: The download tasks above are providing coverage/test data to SonarQube.
So, you should not clean the build

TurpIF:
When creating a pull request, you can create a draft instead. In this way, the team is not requested for review 😉

Amokrane:
> When creating a pull request, you can create a draft instead. In this way, the team is not requested for review 😉

I did want you to review this.

.github/workflows/main.yml
Amokrane: The idea is to try a clean then normal run to see if it would resolve the issue. I wasn't able to find a solution so far, so I am trying some things...

PR #143, opened 2020-10-30
Implement public API collecting ContextData from publisher

TurpIF:
JIRA: EE-1320

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/advancednative/CriteoNativeLoader.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/advancednative/CriteoNativeLoader.java
@@ -28,6 +28,7 @@ import com.criteo.publisher.BidManager;
 import com.criteo.publisher.CriteoErrorCode;
 import com.criteo.publisher.DependencyProvider;
 import com.criteo.publisher.concurrent.RunOnUiThreadExecutor;
+import com.criteo.publisher.context.ContextData;
 import com.criteo.publisher.integration.Integration;
 import com.criteo.publisher.integration.IntegrationRegistry;
 import com.criteo.publisher.model.CdbResponseSlot;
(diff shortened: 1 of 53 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/context/ContextData.java
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Set a new context value in this object
+   *
+   * @param key key-path of an OpenRTB field, or of a Criteo field. Refer to constants in {@link ContextData}
+   * @param value new value for the given key
+   * @return <code>this</code> to chain calls
+   */
+  @Incubating(CONTEXT)
+  public ContextData set(@NonNull String key, @NonNull Iterable<String> value) {
+    // Copy data so it became constant
```
Amokrane, line 68: becomes

publisher-sdk/src/main/java/com/criteo/publisher/advancednative/CriteoNativeLoader.java
```diff
@@ -113,16 +114,16 @@ public View createEmptyNativeView(@NonNull Context context, @Nullable ViewGroup
    */
   public void loadAd() {
     try {
-      doLoad();
+      doLoad(new ContextData());
```
Amokrane: nit: some static factory method such as `none()` would be nice from a readability perspective: `doLoad(ContextData.none())`

publisher-sdk/src/main/java/com/criteo/publisher/context/ContextData.java
```diff
@@ -105,4 +105,20 @@ public ContextData set(@NonNull String key, double value) {
     return Collections.unmodifiableMap(data);
   }
 
+  @Override
+  public boolean equals(Object o) {
```
Amokrane, line 109: Any reason why we are not making this `ContextData` a kotlin data class?

Amokrane (changes requested):
LGTM. Very nitpick comments.

publisher-sdk/src/main/java/com/criteo/publisher/advancednative/CriteoNativeLoader.java
TurpIF: From a `ContextData.none()` I would expect it to be a constant. But this object is mutable so it would be possible to do `ContextData.none().set("toto", "tata")`. Also, this case is kind of special and I don't expect publishers to use it since there is still the API without context.
For test, I plan to elaborate the given context objects so there would be less `new ContextData()` alone.

publisher-sdk/src/main/java/com/criteo/publisher/context/ContextData.java
TurpIF: The `getData()` method is not expected to be used by publisher (package private in Java, or internal in Kotlin). This is a case of object with a mix of mangled/unmangled method. Given the study I made on Kotlin+Java+Proguard interop, we cannot safely use Kotlin here.

Amokrane (approved):

PR #145, opened 2020-11-03
Implement public API collecting UserData from publisher

TurpIF:
JIRA: EE-1319

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/context/EmailHasher.kt
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/context/EmailHasher.kt
@@ -1,0 +1,50 @@
+/*
+ *    Copyright 2020 Criteo
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.criteo.publisher.context
+
+import com.criteo.publisher.annotation.Incubating
+import com.criteo.publisher.annotation.Incubating.CONTEXT
+import java.security.MessageDigest
+import java.util.Locale
+
+object EmailHasher {
+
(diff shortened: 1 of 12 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/context/EmailHasher.kt
```diff
@@ -0,0 +1,50 @@
...
+   * - Hashed with MD5 & output as ASCII text
+   * - Hashed with SHA256 and output as ASCII text
+   */
+  @Incubating(CONTEXT)
+  @JvmStatic
+  fun hash(email: String): String {
+    return email.trim().toLowerCase(Locale.ROOT).toHash("MD5").toHash("SHA-256")
+  }
+
+  private fun String.toHash(type: String): String {
```
Amokrane, line 43: `toHash` might as well be a simple private method. Any advantage in defining it as an extension function?

Amokrane (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/context/UserDataHolder.kt
```diff
@@ -0,0 +1,37 @@
...
+
+package com.criteo.publisher.context
+
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * Indirect reference to the user data to use.
+ *
+ * The [UserData] can be injected by a publisher. So callers of user data should not retain a direct
+ * reference, instead they should always ask this holder for the image loader, as it may change
```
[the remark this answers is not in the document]
TurpIF: Oups, bad copy paste for the image loader. Let me reword this.

Here is a unit test: https://github.com/criteo/android-publisher-sdk/blob/730b2acbca4b2ad7af523b59a2759c0ef191a193/publisher-sdk/src/androidTest/java/com/criteo/publisher/integration/CriteoFunctionalTest.java#L216

Given the API, a publisher is free to change the user data anytime. The user data then needs to be hold somewhere, and guard against concurrent update.

publisher-sdk/src/main/java/com/criteo/publisher/context/EmailHasher.kt
TurpIF: The `trim` and `toLowerCase` are already extension function that make the code readable from left to right. Having `toHash` defined as a simple method would make the code readable first from left to right, then from right to left:

```
toHash(toHash(email.trim().toLowerCase(), "MD5", "SHA-256"))
``` 

IMHO, this is not readable and would require to add local variables.

Note that as the extension is private, it is only local to `EmailHasher`.

Amokrane (approved):

PR #151, opened 2020-11-12
Replace proguard-rules by Keep annotations

Amokrane:


publisher-sdk/src/main/java/com/criteo/publisher/Criteo.java
```diff
@@ -32,10 +32,12 @@
 import com.criteo.publisher.util.DeviceUtil;
 import java.util.List;
 
+@Keep
```
TurpIF, line 35: For your information, `@Keep`ing the whole class means that everything inside is kept. Even private/package elements. Personally, I'm ok to keep them.

This comment is valable for all files

TurpIF: Although, you can remove the `@Keep` on the `setUserData`: https://github.com/criteo/android-publisher-sdk/pull/151/files#diff-ed8fa8b89a351ac154181bc558d65aa2a98aa2b3e4f861c13ffd332f5277025dR186

publisher-sdk/src/main/java/com/criteo/publisher/CriteoAdListener.java
```diff
@@ -16,12 +16,14 @@
...
 package com.criteo.publisher;
 
+import androidx.annotation.Keep;
 import androidx.annotation.NonNull;
 import androidx.annotation.UiThread;
 
 /**
  * All callbacks are invoked on the UI thread, so it is safe to execute any UI operations in the implementation.
  */
+@Keep
```
TurpIF, line 26: I'm not sure if this interface should be kept. Publishers are using the banner/interstitial listeners. They should not care about this interface that is only there to factorize few callbacks.
But I'm not sure if it is fully safe to remove the `@Keep` because of the default. Do you mind testing this without the `@Keep` ? I'm even wondering if we cannot just make this interface package private.

publisher-sdk/proguard-rules.pro
```diff
@@ -42,38 +42,6 @@
 
 # Leaving Criteo SDK models.
```
TurpIF, line 43: As you're working on the proguard-rules, would you mind cleaning it a little bit ? There is a lot of useless rules here.

publisher-sdk/proguard-rules.pro
```diff
@@ -42,38 +42,6 @@
...
 -keepparameternames
```
TurpIF, line 44: In addition to this rule, I would suggest to add few [`-keepattributes`](https://www.guardsquare.com/en/products/proguard/manual/usage#keepattributes). See https://www.guardsquare.com/en/products/proguard/manual/examples#library for instance.

TurpIF (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/CriteoAdListener.java
Amokrane: Removed the `Keep` annotation.  I think making it private or not should be done in a separate PR though.

publisher-sdk/proguard-rules.pro
Amokrane: The retrofit rules? Sure.

publisher-sdk/proguard-rules.pro
Amokrane: Done

publisher-sdk/proguard-rules.pro
```diff
@@ -21,12 +21,9 @@
 #-renamesourcefileattribute SourceFile
 # Retrofit does reflection on generic parameters. InnerClasses is required to use Signature and
```
TurpIF, line 22: Nit: you might update the documentation to indicate that, as an open source project, we keep most info and we obfuscate only to optimize the public API

Also, I guess we can remove all the generated doc above. It doesn't bring any value to this file.

publisher-sdk/proguard-rules.pro
```diff
@@ -37,43 +34,8 @@
 # Guarded by a NoClassDefFoundError try/catch and only used when on the classpath.
 -dontwarn kotlin.Unit
 
-# Top-level functions that can only be used by Kotlin.
--dontwarn retrofit2.-KotlinExtensions
-
 # Leaving Criteo SDK models.
```
TurpIF, line 37: Nit: this comment is not really true now. Also, I wonder if the `keepparamnames` should be above with all the `keepattributes`.

TurpIF (approved):

publisher-sdk/proguard-rules.pro
Amokrane: Done.

TurpIF (approved):

PR #153, opened 2020-11-18
Update picasso to 2.8

Amokrane:
This version of picasso supports AndroidX.

### DIFF
--- a/publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/CriteoImageLoaderTest.java
+++ b/publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/CriteoImageLoaderTest.java
@@ -1,0 +1,55 @@
+/*
+ *    Copyright 2020 Criteo
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.criteo.publisher.advancednative;
+
+import static com.criteo.publisher.StubConstants.STUB_CREATIVE_IMAGE;
+import static com.criteo.publisher.concurrent.ThreadingUtil.callOnMainThreadAndWait;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import android.content.Context;
+import android.widget.ImageView;
+import com.criteo.publisher.mock.MockedDependenciesRule;
+import java.net.URL;
+import javax.inject.Inject;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class CriteoImageLoaderTest {
+
+  @Rule
+  public MockedDependenciesRule mockedDependenciesRule = new MockedDependenciesRule();
+
+  @Inject
+  private Context context;
+
+  @Inject
+  private ImageLoader imageLoader;
+
+  @Test
+  public void givenCreativeImageUrl_ImageLoader_LoadsIt() throws Exception {
+    URL url = new URL(STUB_CREATIVE_IMAGE);
+
+    ImageView imageView = callOnMainThreadAndWait(() -> {
+      ImageView view = new ImageView(context);
+      imageLoader.loadImageInto(url, view, /* placeholder */ null);
+      return view;
+    });
+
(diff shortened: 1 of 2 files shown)


publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/CriteoImageLoaderTest.java
```diff
@@ -0,0 +1,68 @@
...
+        url = new URL("https://placekitten.com/300/500");
+      } catch (MalformedURLException e) {
+        // no-op
+      }
+      assertThat(url).isNotNull();
+      imageLoader.loadImageInto(url, imageView, null);
+      TrackingCommandsExecutor trackingCommandsExecutor = (TrackingCommandsExecutor) dependencyProvider
+          .provideThreadPoolExecutor();
+
+      ThreadingUtil.waitForAllThreads(trackingCommandsExecutor);
```
TurpIF, line 64: I guess that here you cannot use the `mockedDependencyRule.waitForIdleState()` because you are already on the main thread, which is managed. Using the `waitForAllThreads` requires you to cast the thread pool, and implicitly know that idle state is waited this way.

Personally, I prefer to stay mostly on the test thread and jump on UI thread only for precise operations. You have two tools for this:
- `runOnMainThreadAndWait` to execute an action (using `Runnable`)
- `callOnMainThreadAndWait` to execute a function returning a value (using `Callable<V>`)

In this test you need to be twice on the UI thread:
- to instantiate the `ImageView` via `callOnMainThreadAndWait`
- to load the image into the image view via `runOnMainThreadAndWait`

I propose you to use UI thread only for this precise operations. Then you won't need to manually handle exception, manually handle idle state.

TurpIF (changes requested):

publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/CriteoImageLoaderTest.java
```diff
@@ -0,0 +1,68 @@
...
+  public MockedDependenciesRule mockedDependenciesRule = new MockedDependenciesRule();
+
+  @Inject
+  private Context context;
+
+  @Test
+  public void givenKittenUrl_ImageLoader_LoadsIt() throws Exception {
+    ThreadingUtil.runOnMainThreadAndWait(() -> {
+      LayoutInflater inflater = LayoutInflater.from(context);
+      ImageView imageView = (ImageView) inflater.inflate(R.layout.test_image_loader, null);
```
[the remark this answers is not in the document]
Amokrane: Yep, it is possible.

publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/CriteoImageLoaderTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+      DependencyProvider dependencyProvider = mockedDependenciesRule.getDependencyProvider();
+      CriteoImageLoader imageLoader = (CriteoImageLoader) dependencyProvider
```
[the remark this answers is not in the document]
Amokrane: I understand what you are saying, but I have a different point of view. I am not testing the image loader here. I am testing that picasso works well through `CriteoImageLoader` which we know depends on `Picasso`. Testing against an `ImageLoader` doesn't guarantee that `CriteoImageLoader` will be injected. Someone might mock the image loader and set that mock in the `MockedDependencyRule` for instance. OTOH, sure it can also happen that someone mocks `CriteoImageLoader`... I am fine either way, just wanted to explain why I used a `CriteoImageLoader` directly.

publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/CriteoImageLoaderTest.java
Amokrane: Done. That said, `Exception` would have to be handled if I use an `ImageLoader` instead of `CriteoImageLoader` as `ImageLoader` method signatures throw `Exceptions`.

TurpIF: You might use the `Callable` version then. `Callable` throws exception. Then you have to return `null`.

publisher-sdk/src/main/java/com/criteo/publisher/DependencyProvider.java
```diff
@@ -413,7 +413,10 @@ public AdChoiceOverlay provideAdChoiceOverlay() {
 
   @NonNull
   public Picasso providePicasso() {
-    return getOrCreate(Picasso.class, () -> new Picasso.Builder(provideContext()).build());
+    return getOrCreate(
+        Picasso.class,
+        () -> new Picasso.Builder(provideContext()).loggingEnabled(true).build()
```
Amokrane, line 418: left `loggingEnabled` for investigation

TurpIF (approved):

PR #166, opened 2020-12-01
Fix functional test on AppEvent

TurpIF:
This test is broken because of 2aa15017: AsyncTack of AppEvent is
replaced by Executor+Runnable. With AsyncTask, only a single GUM request
can be executed at the same time. With Executor, many GUM requests can
be done in parallel.

With new solution, it appears that two events are sent when starting a
new activity: Launch then Active events.

The former approach is tedious and not deterministic because if network
is slow, then only the "Launch" event is sent, if network is fast, then
both events are sent.

So test is updated to accept the two events. The @FlakyTest is also
removed on this test:
- the new solution is deterministic
- recent work was done to isolate tests, particularly on AppEvents

### DIFF
--- a/publisher-sdk/src/androidTest/java/com/criteo/publisher/integration/CriteoFunctionalTest.java
+++ b/publisher-sdk/src/androidTest/java/com/criteo/publisher/integration/CriteoFunctionalTest.java
@@ -26,6 +26,7 @@ import static org.mockito.AdditionalAnswers.answerVoid;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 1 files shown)


Amokrane:
> With AsyncTask, only a single GUM request
> can be executed at the same time. With Executor, many GUM requests can
> be done in parallel.

Why did the `AsyncTask` ensure that only one call to GUM is made? We were using the `executeOnExecutor` variant which makes it possible to execute tasks in parallel.

TurpIF:
> > With AsyncTask, only a single GUM request
> > can be executed at the same time. With Executor, many GUM requests can
> > be done in parallel.
> 
> Why did the `AsyncTask` ensure that only one call to GUM is made? We were using the `executeOnExecutor` variant which makes it possible to execute tasks in parallel.

There was this condition: https://github.com/criteo/android-publisher-sdk/commit/2aa150178328ad2122c5b140de898300c25ea55e#diff-0d7a200590ded1824cba64fb0b134c84af995b5170fe2961ed52a2a3c2d49564L78

Unfortunately, there was no test on this when we took the project, and the `CriteoFunctionalTest` is tests that we added.

Amokrane (approved):

PR #169, opened 2020-12-04
Tag change (CriteoSdk) + GUM gdpr param key

Amokrane:


publisher-sdk/src/main/java/com/criteo/publisher/logging/ConsoleHandler.kt
```diff
@@ -25,7 +25,7 @@ import com.criteo.publisher.util.BuildConfigWrapper
 internal class ConsoleHandler(private val buildConfigWrapper: BuildConfigWrapper) : LogHandler {
 
   companion object {
-    const val TagPrefix = "Crto"
+    const val TagPrefix = "CriteoSdk"
```
TurpIF, line 28: In the `ConsoleHandler`, could you also substring the full tag so it is not exceeding 23 chars:

> IllegalArgumentException is thrown if the tag.length() > 23 for Nougat (7.0) releases (API <= 23) and prior, there is no tag limit of concern after this API level.

From https://developer.android.com/reference/android/util/Log.html#isLoggable(java.lang.String,%20int)

TurpIF (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/logging/ConsoleHandler.kt
Amokrane: Done

TurpIF (dismissed):

TurpIF (dismissed):

TurpIF (approved):

PR #172, opened 2020-12-08
Remote Handler: Implement sending of logs to CDB

TurpIF:
JIRA: EE-1349

### DIFF
--- a/publisher-sdk/config.groovy
+++ b/publisher-sdk/config.groovy
@@ -79,6 +79,10 @@ adChoiceIconHeightInDp = 15
  * Remote logs configuration
  */
 
+// The batch size of logs sent, at most, in each remote logs request.
+// Given below estimation, a request can have a size of 50KB.
+remoteLogBatchSize = 10
+
 // The relative path in application folder of the sending queue file
 remoteLogQueueFilename = 'criteo_remote_logs_queue'
 
--- a/publisher-sdk/src/main/java/com/criteo/publisher/network/PubSdkApi.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/network/PubSdkApi.java
@@ -21,6 +21,7 @@ import androidx.annotation.Nullable;
 import com.criteo.publisher.csm.MetricRequest;
 import com.criteo.publisher.logging.Logger;
 import com.criteo.publisher.logging.LoggerFactory;
+import com.criteo.publisher.logging.RemoteLogRecords;
 import com.criteo.publisher.model.CdbRequest;
 import com.criteo.publisher.model.CdbResponse;
 import com.criteo.publisher.model.RemoteConfigRequest;
@@ -38,6 +39,7 @@ import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.charset.Charset;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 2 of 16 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/network/PubSdkApi.java
```diff
@@ -134,7 +136,15 @@ public InputStream executeRawGet(URL url) throws IOException {
...
   public void postCsm(@NonNull MetricRequest request) throws IOException {
-    URL url = new URL(buildConfigWrapper.getCdbUrl() + "/csm");
+    postToCdb("/csm", request);
+  }
+
+  public void postLogs(@NonNull List<RemoteLogRecords> request) throws IOException {
+    postToCdb("/inapp/logs", request);
+  }
+
+  private void postToCdb(@NonNull String apiPath, @NonNull Object request) throws IOException {
```
Amokrane, line 146: Do you think it is possible to refactor `loadCdb` to use `postToCdb` as well? (feels a bit weird to have a `postToCdb` method not used by another method that also makes a POST request to CDB).

publisher-sdk/config.groovy
```diff
@@ -79,6 +79,10 @@ adChoiceIconHeightInDp = 15
  * Remote logs configuration
  */
 
+// The batch size of logs sent, at most, in each remote logs requests.
```
Amokrane, line 82: *request

publisher-sdk/config.groovy
```diff
@@ -79,6 +79,10 @@ adChoiceIconHeightInDp = 15
...
+// Given below estimation, request can have a size of 50KB.
```
Amokrane, line 83: *a request

Amokrane (commented):
minor typos (can be done on a separate PR) + open question.

publisher-sdk/src/main/java/com/criteo/publisher/network/PubSdkApi.java
TurpIF: Unfortunately there is few differences between the bid request and the csm/logs request:
- bid request is logged (ok we can give a boolean to control this)
- bid request implies a response, moreover the response cannot just be read via the `JsonSerializer`, it is still using the old `JSONObject` with manual deserialization
- csm/logs does not read the response (just the HTTP status code)

Because of such differences, I don't think we can really use `postToCdb` from `loadCdb`

TurpIF: If we want to factorize the `PubSdkApi`, then `loadConfig` and `loadCdb` are good matches. But again the issue is: logging (not a big issue) and the manual deserialization of `CdbResponse` (bigger issue).
If the deserialization issue is solved, then we could imagine a method like `postToCdbAndRead`.

publisher-sdk/config.groovy
TurpIF: done

publisher-sdk/config.groovy
TurpIF: done

Amokrane (approved):

publisher-sdk/src/main/java/com/criteo/publisher/network/PubSdkApi.java
Amokrane: Ok indeed.

PR #174, opened 2020-12-09
Fix the release workflow for v3.10.2

Amokrane:
action-android is updated as the previous version uses deprecated
commands.

While 3.10.2 is no longer under development, this change is necessary as
this version is needed by a specific publisher.


TurpIF (approved):
LGTM, but don't you need to fix the build too ?

I guess you can cherry pick the commit fixing the CI

Amokrane:
> LGTM, but don't you need to fix the build too ?

Yep I need to fix the build too :)

PR #175, opened 2020-12-09
Bump version of malinskiy's GitHub actions

Amokrane:
This new version solves the deprecated `set-env`.

JIRA: EE-1365


TurpIF (approved):
LGTM, in case of failure, the junit-annotation action will failed, but failure+failure=failure, so we don't care :D

TurpIF (approved):

PR #176, opened 2020-12-09
Make AdvertisingInfo callable on any thread

TurpIF:
The Google API is throwing an exception when called on the main thread:
```
java.lang.IllegalStateException: Calling this from your main thread can lead to deadlock
```

The GAID is fetched in 4 cases:
- when creating the CDB request
- when creating the Bearcat request
- when creating the remote config request
- when creating the remote logs request

The 3 first cases are done on worker thread and current solution is ok.
The remote log case can be done on any thread, including the main one
which is problematic.

To solve this issue we can either:
1. run remote log handler/logger in a (dedicated?) worker thread
2. refactor AdvertisingInfo class to have an async API (listener/future)
3. make AdvertisingInfo callable from any thread

# Proposition 1
Pros:
- slow logger cannot impact business
- no change needed on AdvertisingInfo class
Cons:
- threadId field in logcat or remote log became useless
- impossible or really hard to detect recursive logs and repercussion is
disastrous: instead of 1 stack overflow on 1 thread, the SDK can be a
fork bomb taking 100% of CPU and/or RAM

# Proposition 2
Pros:
- make no assumption
Cons:
- huge refactoring needed
- async API are contaminant: all call stack should be refactored

# Proposition 3
Pros:
- only local change required in AdvertisingInfo class
- no change for caller
Cons:
- assume that GAID is constant
- GAID is available for main thread only after a first call

This commit is implementing the proposition 3.

About the assumption that GAID is constant: to limit the GAID, user
should go deeply in device parameter. When user come back to the
application, either the Android killed the app and the SDK is reinit
again, either the app is still alive, then assumption is false.

About the fact that GAID is available for main thread only after a first
call, this is fixed by prefetching it during SDK init.

### DIFF
--- a/publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/AdvancedNativeFunctionalTest.java
+++ b/publisher-sdk/src/androidTest/java/com/criteo/publisher/advancednative/AdvancedNativeFunctionalTest.java
@@ -41,6 +41,7 @@ import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.clearInvocations;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 import android.content.Intent;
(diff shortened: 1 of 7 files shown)


Amokrane:
The code looks good. That said, I wonder if we could have simply solved this by just letting either the CDB/Bearcat/RemoteConfigRequest fetch the GAID in their respective worker thread  and then just store it in memory so that it can be fetched by the logger at a later time. pro: simplicity. cons: it works as long as those side effects exist.

Amokrane (approved):

TurpIF:
> The code looks good. That said, I wonder if we could have simply solved this by just letting either the CDB/Bearcat/RemoteConfigRequest fetch the GAID in their respective worker thread and then just store it in memory so that it can be fetched by the logger at a later time. pro: simplicity. cons: it works as long as those side effects exist.

Oups, really sorry, I merged without taking a look at your message.

I didn't thought about this solution. Although, it seems to be a degraded mode of the 3rd proposition: without the prefetch and the call on main does not trigger the fetching.

I think this might not solve the initial issue: GAID won't be available in remote logs until an async task happens. Logs can happen before such async task which may throw. So I would say that we should do our best to get the logging part functional ASAP after SDK init.
Please tell me if you have a strong feeling on this, we can open a new PR to correct the merge.

PR #179, opened 2020-12-14
Rephrase log emitted when publisher called SDK init twice

TurpIF:
Log level is downgrade to INFO. Log indicates to dismiss the message if
using mediation adapter (which are calling Criteo#init multiple times)

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/SdkInitLogMessage.kt
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/SdkInitLogMessage.kt
@@ -38,8 +38,11 @@ ${adUnits.joinToString("\n") { "- $it" }}"""
   )
 
   @JvmStatic
-  fun onSdkInitializedMoreThanOnce() = LogMessage(Log.WARN, message =
-    "Criteo SDK initialization method cannot be called more than once"
+  fun onSdkInitializedMoreThanOnce() = LogMessage(message =
+    "Criteo SDK initialization method cannot be called more than once. " +
+        // The mediation adapters always call init before asking for an Ad.
+        // Publishers can't do anything, so they can ignore this log.
+        "Please ignore this if you are using a mediation adapter."
   )
 
   @JvmStatic


publisher-sdk/src/main/java/com/criteo/publisher/SdkInitLogMessage.kt
```diff
@@ -38,8 +38,9 @@ ${adUnits.joinToString("\n") { "- $it" }}"""
   )
 
   @JvmStatic
-  fun onSdkInitializedMoreThanOnce() = LogMessage(Log.WARN, message =
-    "Criteo SDK initialization method cannot be called more than once"
+  fun onSdkInitializedMoreThanOnce() = LogMessage(message =
+    "Criteo SDK initialization method cannot be called more than once. " +
+        "Please ignore this if you're using mediation adapter"
```
Amokrane, line 43: nits: period at the end of the second sentence, `if you are` instead of `if you're` and `a` before mediation adapter.

> Please ignore this if you are using a mediation adapter.

Amokrane: Also, I think we should explain why it is ok to ignore if we are using a mediation adapter.

Amokrane (changes requested):
Submit feedback that must be addressed before merging.

(had to write this as Github doesn't allow me to request changes after pushing single comments without writing something down).

publisher-sdk/src/main/java/com/criteo/publisher/SdkInitLogMessage.kt
TurpIF: > Also, I think we should explain why it is ok to ignore if we are using a mediation adapter. Something like:
> 
> > Please ignore this message if you are using a mediation adapter, as calling init twice in that context is expected.

Publishers do not have to even use the SDK with mediation adapters, they just need to import the gradle dependencies. So I'm not sure that anyone would care about the why, because they don't even know the what (the init).

However, for future maintainers, I can add a comment instead, indicating the why.

Amokrane: Publishers do not have to even use the SDK with mediation adapters, they just need to import the gradle dependencies. So I'm not sure that anyone would care about the why, because they don't even know the what (the init).

I get that, but I don't see why it would make this unnecessary. I am not saying that publishers can call init twice, but that the side effect of init() being called twice as a result of using the mediation adapter, is expected. Also, by printing logs on the console, we now have a leaky abstraction: publishers will know that Criteo SDK is used indirectly.

I am ok with leaving the original message without the why though, just wanted to clarify my thoughts.

Amokrane (approved):

publisher-sdk/src/main/java/com/criteo/publisher/SdkInitLogMessage.kt
TurpIF: Ok, also please note that publishers won't see this log unless they activate the verbose mode.

Amokrane: Alright, ship it!

PR #181, opened 2020-12-14
Fix interstitial test

TurpIF:
The assertion is loosen by accepting `onAdClosed` to be called.

JIRA: EE-1191


Amokrane (approved):

TurpIF:
CI was run several times without error (except MoPub flakyness fixed by #180). So I consider those interstitial tests as fixed.

PR #185, opened 2020-12-22
Send CSM data only if consent is given

Amokrane:
The consent verification is made prior to sending metrics in the sending queue (in other words: what get puts into the sending queue, will be sent).

There are three states:
- CONSENT_GIVEN
- CONSENT_NOT_GIVEN
- CONSENT_UNKNOWN (when the response isn't retrieved from the backend yet).

When we are in the last state, we still write metrics. But, `CONSENT_GIVEN` is mandatory to send metrics to the sending queue.

### DIFF
--- a/publisher-sdk/src/androidTest/java/com/criteo/publisher/csm/CsmFunctionalTest.java
+++ b/publisher-sdk/src/androidTest/java/com/criteo/publisher/csm/CsmFunctionalTest.java
@@ -37,6 +37,7 @@ import androidx.core.util.Consumer;
 import androidx.test.filters.FlakyTest;
 import com.criteo.publisher.Clock;
 import com.criteo.publisher.Criteo;
+import com.criteo.publisher.CriteoInitException;
 import com.criteo.publisher.TestAdUnits;
 import com.criteo.publisher.context.ContextData;
 import com.criteo.publisher.csm.MetricRequest.MetricRequestFeedback;
@@ -86,6 +87,7 @@ public class CsmFunctionalTest {
   public void setUp() throws Exception {
     MockitoAnnotations.initMocks(this);
     integrationRegistry.declare(Integration.IN_HOUSE);
+    givenConsentGiven();
   }
 
   @Test
@@ -137,6 +139,16 @@ public class CsmFunctionalTest {
 
       return true;
     }));
+  }
+
+  /**
+   * Consent status is updated after the first response from CDB. Therefore, a first bid-request
+   * is made here to get the wanted side-effect.
+   */
+  private void givenConsentGiven() throws CriteoInitException {
+    givenInitializedCriteo(TestAdUnits.BANNER_320_480);
+    waitForIdleState();
+    MetricHelper.cleanState(mockedDependenciesRule.getDependencyProvider());
   }
 
   @Test
--- a/publisher-sdk/src/main/java/com/criteo/publisher/csm/CsmBidLifecycleListener.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/csm/CsmBidLifecycleListener.java
@@ -16,6 +16,8 @@
 
 package com.criteo.publisher.csm;
 
+import static com.criteo.publisher.privacy.ConsentData.ConsentStatus.CONSENT_NOT_GIVEN;
+
 import androidx.annotation.NonNull;
 import com.criteo.publisher.Clock;
 import com.criteo.publisher.SafeRunnable;
@@ -27,6 +29,8 @@ import com.criteo.publisher.model.CdbRequestSlot;
 import com.criteo.publisher.model.CdbResponse;
 import com.criteo.publisher.model.CdbResponseSlot;
 import com.criteo.publisher.model.Config;
+import com.criteo.publisher.privacy.ConsentData;
+
 import java.io.InterruptedIOException;
 import java.util.concurrent.Executor;
 
@@ -51,6 +55,9 @@ public class CsmBidLifecycleListener implements BidLifecycleListener {
   private final Config config;
 
   @NonNull
+  private final ConsentData consentData;
+
+  @NonNull
   private final Executor executor;
 
   public CsmBidLifecycleListener(
@@ -58,12 +65,14 @@ public class CsmBidLifecycleListener implements BidLifecycleListener {
       @NonNull MetricSendingQueueProducer sendingQueueProducer,
       @NonNull Clock clock,
       @NonNull Config config,
+      @NonNull ConsentData consentData,
       @NonNull Executor executor
   ) {
     this.repository = repository;
     this.sendingQueueProducer = sendingQueueProducer;
     this.clock = clock;
     this.config = config;
+    this.consentData = consentData;
     this.executor = executor;
   }
 
@@ -287,6 +296,6 @@ public class CsmBidLifecycleListener implements BidLifecycleListener {
   }
 
   private boolean isCsmDisabled() {
-    return !config.isCsmEnabled();
+    return !config.isCsmEnabled() || consentData.getConsentStatus() == CONSENT_NOT_GIVEN;
   }
 }
--- a/publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
@@ -30,18 +30,23 @@ public class CdbResponse {
 
   private static final String TIME_TO_NEXT_CALL = "timeToNextCall";
   private static final String SLOTS = "slots";
+  private static final String CONSENT_GIVEN = "consentGiven";
 
   @NonNull
   private final List<CdbResponseSlot> slots;
 
   private final int timeToNextCall;
 
+  private final boolean consentGiven;
+
   public CdbResponse(
       @NonNull List<CdbResponseSlot> slots,
-      int timeToNextCall
-  ) {
+      int timeToNextCall,
+      @NonNull boolean consentGiven
+      ) {
     this.slots = slots;
     this.timeToNextCall = timeToNextCall;
+    this.consentGiven = consentGiven;
   }
 
   @NonNull
@@ -64,19 +69,28 @@ public class CdbResponse {
       try {
         array = json.getJSONArray(SLOTS);
       } catch (JSONException ex) {
-        logger.debug("Exception while reading slots array" + ex.getMessage());
+        logger.debug("Exception while reading slots array", ex);
       }
       for (int i = 0; i < array.length(); i++) {
         try {
           JSONObject slotStr = array.getJSONObject(i);
           slots.add(CdbResponseSlot.fromJson(slotStr));
         } catch (Exception ex) {
-          logger.debug("Exception while reading slot from slots array" + ex.getMessage());
+          logger.debug("Exception while reading slot from slots array", ex);
         }
       }
     }
 
-    return new CdbResponse(slots, timeToNextCall);
+    boolean consentGiven = false;
+    if (json.has(CONSENT_GIVEN)) {
+      try {
+        consentGiven = json.getBoolean(CONSENT_GIVEN);
+      } catch (JSONException ex) {
+        logger.debug("Exception while reading consentGiven", ex);
+      }
+    }
+
+    return new CdbResponse(slots, timeToNextCall, consentGiven);
   }
 
   @NonNull
@@ -87,6 +101,8 @@ public class CdbResponse {
   public int getTimeToNextCall() {
     return timeToNextCall;
   }
+
+  public boolean getConsentGiven() { return consentGiven; }
 
   @Nullable
   public CdbResponseSlot getSlotByImpressionId(@NonNull String impressionId) {
@@ -104,6 +120,7 @@ public class CdbResponse {
     return "CdbResponse{" +
         "slots=" + slots +
         ", timeToNextCall=" + timeToNextCall +
+        ", consentGiven = " + consentGiven +
         '}';
   }
 }
--- a/publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
@@ -1,0 +1,43 @@
+/*
+ *    Copyright 2020 Criteo
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.criteo.publisher.privacy
+
+import com.criteo.publisher.annotation.OpenForTesting
+import java.util.concurrent.atomic.AtomicReference
+
+@OpenForTesting
+class ConsentData {
+
+  /**
+   * Consent can be in any of the following two states:
+   *
+   * <ul>
+   *  <li> CONSENT_GIVEN: it is known that consent has been given by the user</li>
+   *  <li> CONSENT_NOT_GIVEN: it is known that consent has not been given by the user,
+   *  or isn't known yet.</li>
+   * </ul>
+   */
+  enum class ConsentStatus {
+    CONSENT_GIVEN,
+    CONSENT_NOT_GIVEN
+  }
+  private val valueRef = AtomicReference<ConsentStatus>(ConsentStatus.CONSENT_NOT_GIVEN)
+
+  var consentStatus: ConsentStatus
+    get() = valueRef.get()
+    set(value) = valueRef.set(value)
+}
--- a/publisher-sdk/src/test/java/com/criteo/publisher/LiveCdbCallListenerTests.kt
+++ b/publisher-sdk/src/test/java/com/criteo/publisher/LiveCdbCallListenerTests.kt
@@ -21,6 +21,7 @@ import com.criteo.publisher.model.CacheAdUnit
 import com.criteo.publisher.model.CdbRequest
 import com.criteo.publisher.model.CdbResponse
 import com.criteo.publisher.model.CdbResponseSlot
+import com.criteo.publisher.privacy.ConsentData
 import com.nhaarman.mockitokotlin2.any
 import com.nhaarman.mockitokotlin2.never
 import com.nhaarman.mockitokotlin2.times
@@ -44,6 +45,9 @@ class LiveCdbCallListenerTests {
 
   @Mock
   private lateinit var cacheAdUnit: CacheAdUnit
+
+  @Mock
+  private lateinit var consentData: ConsentData
 
   @InjectMocks
   private lateinit var liveCdbCallListener: LiveCdbCallListener
@@ -181,5 +185,23 @@ class LiveCdbCallListenerTests {
     verify(bidLifecycleListener).onCdbCallFailed(cdbRequest, exception)
     verify(bidLifecycleListener, never()).onBidConsumed(any(), any())
     verify(bidLifecycleListener).onCdbCallFailed(cdbRequest, exception)
+  }
+
+  @Test
+  fun onBidResponse_givenConsentGiven_ThenUpdateConsentDataAccordingly() {
+    whenever(cdbResponse.consentGiven).thenReturn(true)
+
+    liveCdbCallListener.onCdbResponse(cdbRequest, cdbResponse)
+
+    verify(consentData).consentStatus = ConsentData.ConsentStatus.CONSENT_GIVEN
+  }
+
+  @Test
+  fun onBidResponse_givenConsentNotGiven_ThenUpdateConsentDataAccordingly() {
+    whenever(cdbResponse.consentGiven).thenReturn(false)
+
+    liveCdbCallListener.onCdbResponse(cdbRequest, cdbResponse)
+
+    verify(consentData).consentStatus = ConsentData.ConsentStatus.CONSENT_NOT_GIVEN
   }
 }
--- a/publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
+++ b/publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
@@ -24,11 +24,13 @@ import com.criteo.publisher.model.CdbRequestSlot
 import com.criteo.publisher.model.CdbResponse
 import com.criteo.publisher.model.CdbResponseSlot
 import com.criteo.publisher.model.Config
+import com.criteo.publisher.privacy.ConsentData
+import com.criteo.publisher.privacy.ConsentData.ConsentStatus.CONSENT_GIVEN
+import com.criteo.publisher.privacy.ConsentData.ConsentStatus.CONSENT_NOT_GIVEN
 import com.criteo.publisher.util.AdUnitType.CRITEO_BANNER
 import com.nhaarman.mockitokotlin2.any
 import com.nhaarman.mockitokotlin2.argumentCaptor
 import com.nhaarman.mockitokotlin2.check
-import com.nhaarman.mockitokotlin2.doAnswer
 import com.nhaarman.mockitokotlin2.doReturn
 import com.nhaarman.mockitokotlin2.eq
 import com.nhaarman.mockitokotlin2.mock
@@ -63,6 +65,9 @@ class CsmBidLifecycleListenerTest {
   @Mock
   private lateinit var config: Config
 
+  @Mock
+  private lateinit var consentData: ConsentData
+
   private lateinit var executor: Executor
 
   private lateinit var listener: CsmBidLifecycleListener
@@ -77,13 +82,23 @@ class CsmBidLifecycleListenerTest {
       on { isCsmEnabled } doReturn true
     }
 
+    whenever(consentData.consentStatus).thenReturn(CONSENT_GIVEN)
+
     listener = CsmBidLifecycleListener(
         repository,
         sendingQueueProducer,
         clock,
         config,
+        consentData,
         executor
     )
+  }
+
+  @Test
+  fun onSdkInitialized_PushAllMetricsInQueue() {
+    listener.onSdkInitialized()
+
+    verify(sendingQueueProducer).pushAllInQueue(repository)
   }
 
   @Test
@@ -96,15 +111,26 @@ class CsmBidLifecycleListenerTest {
   }
 
   @Test
-  fun onSdkInitialized_PushAllMetricsInQueue() {
+  fun onSdkInitialized_GivenConsentNotGiven_DoNothing() {
+    givenConsentNotGiven()
+
     listener.onSdkInitialized()
 
-    verify(sendingQueueProducer).pushAllInQueue(repository)
+    verifyFeatureIsDeactivated()
   }
 
   @Test
   fun onCdbCallStarted_GivenDeactivatedFeature_DoNothing() {
     givenDeactivatedFeature()
+
+    listener.onCdbCallStarted(mock())
+
+    verifyFeatureIsDeactivated()
+  }
+
+  @Test
+  fun onCdbCallStarted_GivenConsentNotGiven_DoNothing() {
+    givenConsentNotGiven()
 
     listener.onCdbCallStarted(mock())
 
@@ -139,7 +165,7 @@ class CsmBidLifecycleListenerTest {
 
     verifyFeatureIsDeactivated()
   }
-
+  
   @Test
   fun onCdbCallFinished_GivenOnlyNoBid_PushReadyToSendInQueue() {
     val request = givenCdbRequestWithSlots("id")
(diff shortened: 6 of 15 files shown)


Amokrane:
WIP:
1. Need to fix the SONAR warnings.
2. In the scenario where consent is not given then given (before a metric is marked as `readyToSend`),  we will end up in a case where metric files will be half-written. I think that all CsmBidLifecycle callbacks, except the first one (CdbCallStarted), should not create a new file if it doesn't exist already.

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
```diff
@@ -76,7 +81,16 @@ public static CdbResponse fromJson(@NonNull JSONObject json) {
       }
     }
 
-    return new CdbResponse(slots, timeToNextCall);
+    boolean consentGiven = false;
+    if (json.has(CONSENT_GIVEN)) {
+      try {
+        consentGiven = json.getBoolean(CONSENT_GIVEN);
+      } catch (JSONException ex) {
+        logger.debug("Exception while reading consentGiven" + ex.getMessage());
```
TurpIF, line 89: You should not swallow an exception. Please log the whole exception:
```
logger.debug("message", ex);
```

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -0,0 +1,46 @@
...
+   * <ul>
+   *  <li> CONSENT_GIVEN: it is known that consent has been given by the user</li>
+   *  <li> CONSENT_NOT_GIVEN: it is known that consent has not been given by the user</li>
+   *  <li> CONSENT_UNKNOWN: consent is not known yet (pending CDB request) </li>
+   * </ul>
+   */
+  enum class ConsentStatus {
+    CONSENT_GIVEN,
+    CONSENT_NOT_GIVEN,
+    CONSENT_UNKNOWN
```
TurpIF, line 37: As told during the standup, we can drop the the unknown part and just keep a boolean.
The unknown correspond to no consent given.

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
```diff
@@ -76,7 +81,16 @@ public static CdbResponse fromJson(@NonNull JSONObject json) {
       }
     }
 
-    return new CdbResponse(slots, timeToNextCall);
+    boolean consentGiven = false;
```
TurpIF, line 84: By having a default value to `false` instead of `null`, you should be sure that backend will always returns the consent even in case of no bid.

I'm ok with that, I made a ticket to check and implement this on backend: EE-1401

TurpIF, line 84: By having a default value to `false` instead of `null`, you should be sure that backend will always returns the consent even in case of no bid.

I'm ok with that, I made a ticket to check and implement this on backend: EE-1401

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
TurpIF: Btw, maybe could you apply this suggestion to the whole file ?

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
```diff
@@ -30,18 +30,23 @@
 
   private static final String TIME_TO_NEXT_CALL = "timeToNextCall";
   private static final String SLOTS = "slots";
+  private static final String CONSENT_GIVEN = "consentGiven";
```
TurpIF, line 33: What do you think about transforming this class into an Java `@AutoValue` class or a Kotlin `data` class ?

Deserialization is isolating potential error per slots, but except the prefetch (which might be removed), all bid response contains only a single slot.

test-utils/src/main/java/com/criteo/publisher/TestDeviceIds.kt
```diff
@@ -0,0 +1,19 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.criteo.publisher
+
+const val NON_CONSENTED_DEVICE_ID = "2"
```
TurpIF, line 19: Nit: the project already contains the `StubConstants` class containing constants.
I think that this file will not contains anything but this line. What do you think about moving this into `StubConstants` instead ?

TurpIF (changes requested):
As told during the stand up, functional change is requested:
- SDK only treat consent as a boolean value (given or not)
- By default, consent is not given
- Consent is cached in shared pref
- For each CDB requests, consent is updated (this implies an accepted lag of 1 bid)
- If consent is not given, then CSM and remote log are disabled
- (1) To determine: in case of transition from consent given to consent not given, what should we do on persisted CSM/logs:
  - Clean persisted data: more work
  - Do nothing: consent was given when gathering such data
- (2) To determine: in case of transition from consent not given to consent given, should we accept partially filled CSM metrics ?
  - Yes: no work to do, but partial metrics
  - No: add a lot of complexity, for always filled metrics

On (1) I would tend to do nothing: it is edge case adding complexity.
On (2) I would tend to accept partial metrics. Partial metrics can already occurs (in case of thrown exception for instance). The transition is an edge case so expected volume of such event is low. We can react later if we see issues on CSM.

test-utils/src/main/java/com/criteo/publisher/TestDeviceIds.kt
Amokrane: Well, StubConstants contains .. stub constants which isn't the case of this deviceId. I feel that even if there is only a single constant in this file, it is still worth separating concerns and having clear boundaries. Or we should rename `StubConstants` into `TestConstants` or something like that.

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
Amokrane: I think this is a good idea, but I don't think we should do this in this PR. It should be done separately.

test-utils/src/main/java/com/criteo/publisher/TestDeviceIds.kt
TurpIF: Renaming `StubConstants` will be a breaking change in lot of tests. Included ones of mediation adapters.
Ok to keep this file then. Maybe should you encapsulate the constant in an kotlin `object` or use `@file:JvmName` to make it nicer from Java code (see https://stackoverflow.com/a/57355475/3663982 to make your choice).

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
Amokrane: I am ok with dropping the `CONSENT_UNKNOWN` constant. Note that we are loosing the ability to distinguish between `CONSENT_NOW_GIVEN` and `CONSENT_UNKNOWN` in console logs, in this case.

publisher-sdk/src/main/java/com/criteo/publisher/model/CdbResponse.java
Amokrane: Yeah was gonna say that I just copy/pasted whatever was done above :) (not that it's a good excuse).

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -0,0 +1,46 @@
...
+  }
+  private val valueRef = AtomicReference<ConsentStatus>(ConsentStatus.CONSENT_UNKNOWN)
+
+  fun getConsentStatus(): ConsentStatus = valueRef.get()
```
TurpIF, line 41: Nit: this can be a kotlin `var` with `get()/set()`

Amokrane:
> As told during the stand up, functional change is requested:
> 
> * SDK only treat consent as a boolean value (given or not)
> * By default, consent is not given
> * Consent is cached in shared pref
> * For each CDB requests, consent is updated (this implies an accepted lag of 1 bid)
> * If consent is not given, then CSM and remote log are disabled
> * (1) To determine: in case of transition from consent given to consent not given, what should we do on persisted CSM/logs:
>   
>   * Clean persisted data: more work
>   * Do nothing: consent was given when gathering such data
> * (2) To determine: in case of transition from consent not given to consent given, should we accept partially filled CSM metrics ?
>   
>   * Yes: no work to do, but partial metrics
>   * No: add a lot of complexity, for always filled metrics
> 
> On (1) I would tend to do nothing: it is edge case adding complexity.
> On (2) I would tend to accept partial metrics. Partial metrics can already occurs (in case of thrown exception for instance). The transition is an edge case so expected volume of such event is low. We can react later if we see issues on CSM.


> To determine: in case of transition from consent given to consent not given, what should we do on persisted CSM/logs:

For that transition, we also need to make sure that we don't further write any CSM/logs after the transition. If we don't do anything, then we will continue writing new CSMs/logs during the init phase of the SDK (and until we receive a response from Cdb). I think we should monitor if any of the CMP values have changed on the client side and change the consent back to `CONSENT_NOT_GIVEN` in that case. This can be done either during init only (by keeping an old copy of the CMP data) or by registering callbacks. WDYT?

publisher-sdk/src/androidTest/java/com/criteo/publisher/csm/CsmFunctionalTest.java
```diff
@@ -59,14 +66,28 @@
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
 import org.mockito.MockitoAnnotations;
 
 @FlakyTest(detail = "CSM are triggered concurrently of bid")
+@RunWith(Parameterized.class)
```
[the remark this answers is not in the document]
Amokrane: Good to know, I'll check.

TurpIF:
> > * (1) To determine: in case of transition from consent given to consent not given, what should we do on persisted CSM/logs:
> >   
> >   * Clean persisted data: more work
> >   * Do nothing: consent was given when gathering such data
> > * (2) To determine: in case of transition from consent not given to consent given, should we accept partially filled CSM metrics ?
> >   
> >   * Yes: no work to do, but partial metrics
> >   * No: add a lot of complexity, for always filled metrics
> > 
> > On (1) I would tend to do nothing: it is edge case adding complexity.
> > On (2) I would tend to accept partial metrics. Partial metrics can already occurs (in case of thrown exception for instance). The transition is an edge case so expected volume of such event is low. We can react later if we see issues on CSM.
> 
> Ok for both transitions.
> 
> However, for:
> 
> > To determine: in case of transition from consent given to consent not given, what should we do on persisted CSM/logs:
> 
> For that transition, we also need to make sure that we don't further write any CSM/logs after the transition. If we don't do anything, then we will continue writing new CSMs/logs during the init phase of the SDK (and until we receive a response from Cdb). I think we should monitor if any of the CMP values have changed on the client side and change the consent back to `CONSENT_NOT_GIVEN` in that case. This can be done either during init only (by keeping an old copy of the CMP data) or by registering callbacks. WDYT?

By transition, I meant transition in backend response: so SDK is just sending the consent strings to the backend and listening the response. So the consent strings are never read by the SDK.

There you're talking about transition in the consent strings. Which is 1 bid ahead. But I thought that we accepted this delay:

> For each CDB requests, consent is updated (this implies an accepted lag of 1 bid)

Please note that the delay is symmetrical. It is not to bias the decision in any way. This is to truly reflect the backend decision.

Moreover, keeping a state of consent strings (GDPR, CCPA, MoPub, iOS, ...) and invalidating consent based on this can:
- at best: do its job with the cost of technical complexity
- at worst: kill CSM or remote logging features because consent strings may legitimately change: for instance if they contains a timestamp, ...)

Adding the fact that transitions are uncommon edge cases (I really never saw any app asking twice a consent), I would prefer to only handle transitions in backend response, not transitions in consent strings.

Amokrane:
> By transition, I meant transition in backend response: so SDK is just sending the consent strings to the backend and listening the response. So the consent strings are never read by the SDK.

I am not sure I get this point: `So the consent strings are never read by the SDK.`. It is the SDK that reads the consent strings, not the backend. reading != interpreting or treating.

> Adding the fact that transitions are uncommon edge cases (I really never saw any app asking twice a consent), I would prefer 
to only handle transitions in backend response, not transitions in consent strings.

It's not that the app will ask the consent twice, but users can and (will) modify their consent by going to the device or app settings. 
See the screen below, taken from LeMonde application, where I can modify my consent.
Webp net-resizeimage (1)

If I remove my consent when will the app take into account the new consent and stop sending CSMs/logs? With the current implementation, it will happen after the next Cdb response and during the time window before that next Cdb response all my data will still be logged which does not really comply with legislation I suppose :)

> There you're talking about transition in the consent strings. Which is 1 bid ahead. But I thought that we accepted this delay:

It is acceptable if we are missing logs in the case where we don't yet that consent was given. Whether it is  acceptable to log data that we shouldn't log, (even for a 1-bid lag) shouldn't be decided by us IMHO, let's double check with PM on this.

> at best: do its job with the cost of technical complexity
at worst: kill CSM or remote logging features because consent strings may legitimately change: for instance if they contains a timestamp, ...)

Consent data don't have any timestamp and won't change unless the user has changed them, or unless there is a new TCF/CCPA version.  Technical complexity: well it's not that complex to listen for SharedPref changes or keep a copy of the old data.

I am ok to go with the simple solution, but we should get an explicit agreement over this from PM.

TurpIF:
> > By transition, I meant transition in backend response: so SDK is just sending the consent strings to the backend and listening the response. So the consent strings are never read by the SDK.
> 
> I am not sure I get this point: `So the consent strings are never read by the SDK.`. It is the SDK that reads the consent strings, not the backend. reading != interpreting or treating.

Yep sorry, I wrongly express myself. Instead of `read`, I meant `used` (except for forging the bid request).
 
> > Adding the fact that transitions are uncommon edge cases (I really never saw any app asking twice a consent), I would prefer
> > to only handle transitions in backend response, not transitions in consent strings.
> 
> It's not that the app will ask the consent twice, but users can (and will) modify their consent by going to the device or app settings.
> See the screen below, taken from LeMonde application, where I can modify my consent.
> Webp net-resizeimage (1)

To be honest, that's the first time I see a screen like that.

> I am ok to go with the simple solution, but we should get an explicit agreement over this from PM.

Ok, let's ask to him next week. Meanwhile, I think that setting up listener+consent invalidation can be done on top of this PR. What do you think to move on and add a new ticket as a reminder ?

Amokrane:
> Ok, let's ask to him next week. Meanwhile, I think that setting up listener+consent invalidation can be done on top of this PR. What do you think to move on and add a new ticket as a reminder ?

Sounds good to me, let's do that.

publisher-sdk/src/androidTest/java/com/criteo/publisher/csm/CsmFunctionalTest.java
```diff
@@ -59,14 +66,28 @@
...
+  @Parameterized.Parameters
+  public static Collection consents() {
+    return Arrays.asList(new Object[][] {
+        { true },
+        { false }
+    });
+  }
+
+  @Parameter(0)
+  public boolean consentGiven;
```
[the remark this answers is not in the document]
Amokrane: After working again on the change I am now thinking that these functional tests should only cover the nominal case: when CSM is enabled. The case where CSM is disabled (either due to consent or CaC which by the way isn't covered here) is already covered in unit tests.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
Amokrane: If I do that then I would need to expose the `AtomicReference`. I am finding it cleaner to expose the value directly. If I don't use an `AtomicReference`, then I would need to make it volatile. (the consent status is written/read on different threads within code that is not synchronized).

TurpIF: The `valueRef` can stay `private` and a `var` could use it:
```
private val valueRef = atomic

var consentStatus: ConsentStatus
  get() = valueRef.get()
  set(value) = valueRef.set(value)
```

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -0,0 +1,43 @@
...
+package com.criteo.publisher.privacy
+
+import com.criteo.publisher.annotation.OpenForTesting
+import java.util.concurrent.atomic.AtomicReference
+
+@OpenForTesting
+class ConsentData {
+
+  /**
+   * Consent can be in any of the following three states:
```
TurpIF, line 26: `s/three/two/`

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * <ul>
+   *  <li> CONSENT_GIVEN: it is known that consent has been given by the user</li>
+   *  <li> CONSENT_NOT_GIVEN: it is known that consent has not been given by the user,
+   *  or isn't known yet.</li>
+   * </ul>
+   */
+  enum class ConsentStatus {
```
TurpIF, line 34: Do you have a reason to use an enum value rather than a simple boolean ?
Using enum makes implementation a little bit harder to get.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -0,0 +1,43 @@
...
+    CONSENT_GIVEN,
+    CONSENT_NOT_GIVEN
+  }
+  private val valueRef = AtomicReference<ConsentStatus>(ConsentStatus.CONSENT_NOT_GIVEN)
```
TurpIF, line 38: This value is not persisted. So you're always losing the consent at each new session. 
Do you plan to persist it in another PR ?

publisher-sdk/src/test/java/com/criteo/publisher/LiveCdbCallListenerTests.kt
```diff
@@ -182,4 +186,22 @@ class LiveCdbCallListenerTests {
     verify(bidLifecycleListener, never()).onBidConsumed(any(), any())
     verify(bidLifecycleListener).onCdbCallFailed(cdbRequest, exception)
   }
+
+  @Test
+  fun onBidResponse_givenConsentGiven_ThenUpdatConsentDataAccordingly() {
```
TurpIF, line 191: s/Updat/Update

publisher-sdk/src/test/java/com/criteo/publisher/LiveCdbCallListenerTests.kt
```diff
@@ -182,4 +186,22 @@ class LiveCdbCallListenerTests {
...
+    whenever(cdbResponse.consentGiven).thenReturn(true)
+
+    liveCdbCallListener.onCdbResponse(cdbRequest, cdbResponse)
+
+    verify(consentData).consentStatus = ConsentData.ConsentStatus.CONSENT_GIVEN
+  }
+
+  @Test
+  fun onBidResponse_givenConsentNotGiven_ThenUpdatConsentDataAccordingly() {
```
TurpIF, line 200: s/Updat/Update

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
```diff
@@ -96,10 +104,12 @@ class CsmBidLifecycleListenerTest {
   }
 
   @Test
-  fun onSdkInitialized_PushAllMetricsInQueue() {
+  fun onSdkInitialized_GivenConsentNotGiven_DoNothing() {
```
TurpIF, line 107: There is no more test covering the normal case: on SDK init, CSM are pushed in the queue

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
```diff
@@ -284,8 +257,20 @@ class CsmBidLifecycleListenerTest {
...
+
+    listener.onBidConsumed(adUnit, mock())
+
+    verifyFeatureIsDeactivated()
+  }
+
+
   @Test
   fun onBidConsumed_GivenNotExpiredBid_SetElapsedTimeAndReadyToSend() {
+    givenConsentGiven()
```
TurpIF, line 273: Nit: not useful as your setup already gives the consent

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
```diff
@@ -303,12 +288,11 @@ class CsmBidLifecycleListenerTest {
       verify(it).setElapsedTimestamp(42)
       verify(it).setReadyToSend(true)
     }
-
-    verify(sendingQueueProducer).pushInQueue(repository, "id")
   }
 
   @Test
   fun onBidConsumed_GivenExpiredBid_SetReadyToSend() {
+    givenConsentGiven()
```
TurpIF, line 295: Nit: not useful as your setup already gives the consent

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
```diff
@@ -141,54 +161,8 @@ class CsmBidLifecycleListenerTest {
...
-
-    listener.onCdbCallFinished(request, response)
-
-    verify(sendingQueueProducer, never()).pushInQueue(any(), any())
-  }
-
-  @Test
-  fun onCdbCallFinished_GivenNoBidAndInvalidBidAndValidBidReceived_UpdateThemByIdAccordingly() {
+  fun onCdbCallFinished_GivenConsentGivenAndNoBidAndInvalidBidAndValidBidReceived_UpdateThemByIdAccordingly() {
+    givenConsentGiven()
```
TurpIF, line 165: Nit: not useful as your setup already gives the consent

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
```diff
@@ -132,7 +151,8 @@ class CsmBidLifecycleListenerTest {
   }
 
   @Test
-  fun onCdbCallFinished_GivenDeactivatedFeature_DoNothing() {
+  fun onCdbCallFinished_GivenConsentGivenAndDeactivatedFeature_DoNothing() {
+    givenConsentGiven()
```
TurpIF, line 155: Nit: not useful as your setup already gives the consent

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
```diff
@@ -338,6 +320,24 @@ class CsmBidLifecycleListenerTest {
     verifyZeroInteractions(repository)
   }
 
+  @Test
+  fun onBidsCached_GivenDeactivatedFeature_DoNothing() {
+   givenDeactivatedFeature()
```
TurpIF, line 325: There is a missing space here

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
```diff
@@ -141,54 +161,8 @@ class CsmBidLifecycleListenerTest {
   }
 
   @Test
-  fun onCdbCallFinished_GivenOnlyNoBid_PushReadyToSendInQueue() {
```
TurpIF, line 144: Why are all those tests/assertions removed ?

test-utils/src/main/java/com/criteo/publisher/network/CdbMock.kt
```diff
@@ -141,9 +141,11 @@ class CdbMock(private val jsonSerializer: JsonSerializer) {
 
       val responseSlots = cdbRequest.slots.mapNotNull { it.toResponseSlot() }.joinToString()
       val requestId = UUID.randomUUID().toString()
+
       val cdbResponse = """
-      {
+      {z
```
TurpIF, line 146: There is an extra `z`

TurpIF (changes requested):

publisher-sdk/src/androidTest/java/com/criteo/publisher/csm/CsmFunctionalTest.java
```diff
@@ -404,4 +430,11 @@ private void waitForIdleState() {
   private void loadBid(@NonNull AdUnit adUnit) {
     Criteo.getInstance().loadBid(adUnit, new ContextData(), ignored -> { /* no op */ });
   }
+
+  private void triggerMetricRequest() {
+    threadPoolExecutor.execute(() -> metricSendingQueueProducer.pushAllInQueue(metricRepository));
```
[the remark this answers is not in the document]
Amokrane: Ah I didn't know about `MetricHelper`, will do.

publisher-sdk/src/main/java/com/criteo/publisher/csm/CsmBidLifecycleListener.java
```diff
@@ -287,6 +290,6 @@ private void updateByCdbRequestIds(@NonNull CdbRequest request, @NonNull MetricU
   }
 
   private boolean isCsmDisabled() {
-    return !config.isCsmEnabled();
+    return !config.isCsmEnabled() || consentData.getConsentStatus().equals(CONSENT_NOT_GIVEN);
```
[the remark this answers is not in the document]
Amokrane: Yep sure.

publisher-sdk/src/main/java/com/criteo/publisher/csm/CsmBidLifecycleListener.java
```diff
@@ -58,21 +65,17 @@ public CsmBidLifecycleListener(
...
   ) {
     this.repository = repository;
     this.sendingQueueProducer = sendingQueueProducer;
     this.clock = clock;
     this.config = config;
+    this.consentData = consentData;
     this.executor = executor;
   }
 
-  /**
```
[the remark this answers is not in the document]
Amokrane: Removed it when I wrote the first implementation where onSdkinit was no longer doing anything. Putting the comment back.

publisher-sdk/src/main/java/com/criteo/publisher/csm/MetricDirectory.java
```diff
@@ -105,7 +106,6 @@ private String getMetricFilenameFromImpressionId(@NonNull String impressionId) {
    * This is the dual of {@link #getMetricFilenameFromImpressionId(String)}: The impression ID is a
    * string, and the metric filename is the impression ID with an extension. To find again the
    * impression ID from the filename, then this methods only have to remove the extension.
-   *
```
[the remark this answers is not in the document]
Amokrane: Yeah, actually I added code with the previous implementation which I removed back again.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
Amokrane: I feel it is cleaner and makes the code easier to read. It's unclear seeing `true/false` being passed around vs. seeing values such as these ones.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
Amokrane: Yep planned for another PR.

publisher-sdk/src/test/java/com/criteo/publisher/csm/CsmBidLifecycleListenerTest.kt
Amokrane: Good catch, I removed it back when I wrote the first implementation.

publisher-sdk/src/androidTest/java/com/criteo/publisher/csm/CsmFunctionalTest.java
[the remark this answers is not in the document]
Amokrane: Though not gonna lie, it is an `enum` because initially there were three states. I can change it to `boolean` in the next PR where the consent is stored.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
TurpIF: It might be cleaner internally to the `ConsentData` class. Although, even if there was a third state, I would argue that callers should not be able to know about internal state and should only consume a boolean decision: do I have the consent or not ?

Amokrane: Yeah I will move it to boolean in another PR.

Amokrane:
I am fixing the code smells.

Amokrane:
Actually, to minimize build and waiting times, I would rather fix the minor code smells on the other PR.

TurpIF (approved):
Ok for doing the minor fix in another PR

PR #187, opened 2020-12-29
Make ConsentStatus value a Boolean instead of Enum + Persist ConsentData

Amokrane:


publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -16,28 +16,22 @@
...
+class ConsentData(val sharedPreferences: SharedPreferences) {
+  companion object {
+    const val CRITEO_CONSENT_GIVEN_KEY = "CRTO_ConsentGiven"
   }
-  private val valueRef = AtomicReference<ConsentStatus>(ConsentStatus.CONSENT_NOT_GIVEN)
 
-  var consentStatus: ConsentStatus
-    get() = valueRef.get()
-    set(value) = valueRef.set(value)
+  @Synchronized
```
TurpIF, line 28: Why do you need those methods to be synchronized ?

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -16,28 +16,22 @@
...
+  fun getConsentGiven() = sharedPreferences.getBoolean(CRITEO_CONSENT_GIVEN_KEY, false)
```
TurpIF, line 29: What do you think about using the `is` prefix instead (`isConsentGiven()`) ?

publisher-sdk/src/androidTest/java/com/criteo/publisher/privacy/ConsentDataTest.kt
```diff
@@ -0,0 +1,59 @@
...
+  val mockedDependenciesRule = MockedDependenciesRule()
+
+  @Inject
+  private lateinit var context: Context
+
+  private lateinit var sharedPreferences: SharedPreferences
+
+  @Before
+  fun setUp() {
+    sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
```
TurpIF, line 42: The shared pref injected there https://github.com/criteo/android-publisher-sdk/pull/187/files#diff-488c51c5458586a79ec9705c7ee3d77c8b2a5b5b88dbb9879e78148f257ff704R712 is the private shared pref of the SDK.
I guess that using the SDK one is better than using the default one, to avoid colliding with the publisher.

Also, when using the SDK one in tests, you can simply put an `@Inject` annotation to get it. Also, you don't need to clean its state, the `MockedDependenciesRule` already handle it.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
```diff
@@ -16,28 +16,22 @@
...
-   *  <li> CONSENT_NOT_GIVEN: it is known that consent has not been given by the user,
-   *  or isn't known yet.</li>
-   * </ul>
-   */
-  enum class ConsentStatus {
-    CONSENT_GIVEN,
-    CONSENT_NOT_GIVEN
+class ConsentData(val sharedPreferences: SharedPreferences) {
+  companion object {
+    const val CRITEO_CONSENT_GIVEN_KEY = "CRTO_ConsentGiven"
```
TurpIF, line 25: either the companion or the constant should be private

publisher-sdk/src/androidTest/java/com/criteo/publisher/privacy/ConsentDataTest.kt
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @After
+  fun tearDown() {
+    sharedPreferences.edit().clear().commit()
+  }
+
+  @Test
```
TurpIF, line 50: Would you mind adding a test to check the behavior when the shared pref is empty ?

TurpIF (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
Amokrane: read & writes happening on different threads & unsynchronized code.

publisher-sdk/src/androidTest/java/com/criteo/publisher/privacy/ConsentDataTest.kt
Amokrane: > I guess that using the SDK one is better than using the default one, to avoid colliding with the publisher.

Yes, and in the public code it is the private one that is used not the public one.

But sure I can use the private one in the test as well.

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
TurpIF: From the `SharedPreference` javadoc:

> Note: This class provides strong consistency guarantees. It is using expensive operations which might slow down an app. Frequently changing properties or properties where loss can be tolerated should use other mechanisms. For more details read the comments on SharedPreferences.Editor.commit() and SharedPreferences.Editor.apply().

And for `SharedPreferences.Editor.apply()`:

> This atomically performs the requested modifications, replacing whatever is currently in the SharedPreferences.

Both reads are consistent and writes are atomic. So you don't need to synchronize the access even when used from different threads.

TurpIF (approved):
Nitpicking on synchronization

publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
Amokrane: Yes, but if I want to read the latest available value then it is helpful to synchronize (since writing involves grabbing the `editor` reference first before calling `apply`). Though I agree it's not mandatory. I can remove.

PR #188, opened 2020-12-30
Remove the synchronization in ConsentData

Amokrane:
This will avoid unnecessary contention. `apply` is an atomic operation, but writing 
involves grabbing the reference to the `editor` first. Therefore, the only effect is that 
reading might not return the latest available value, but that is an acceptable trade-off.

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/privacy/ConsentData.kt
@@ -25,10 +25,8 @@ class ConsentData(val sharedPreferences: SharedPreferences) {
     private const val CRITEO_CONSENT_GIVEN_KEY = "CRTO_ConsentGiven"
   }
 
-  @Synchronized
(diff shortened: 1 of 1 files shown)


TurpIF (approved):
Note that synchronizing at a such low level is done to respect consistency of internal values. Here the shared pref is already consistent, so the sync does not help to anything.

If you want to guarantee an order, then caller should be synchronized first, not the callee.

For instance the code below does not guarantee any order even with `@Synchronized`:
```
val executor = Executors.newFixedThreadPool(2)

executor.submit {
    // <-- The OS is free to do a context switch here, making the second code below run before this one.
    consentData.setConsentGiven(true)
}

executor.submit {
    consentData.isConsentGiven()
}
```

Amokrane:
Sure, but I didn't say that it was for ordering purposes at the caller level. The reason why I put the `Synchronized` was to make sure that when `setConsentData` is being executed, then it won't let a `read` operation execute before `setContentData` returns.

PR #190, opened 2021-01-04
Deactivate remote logs when consent is not given

Amokrane:


publisher-sdk/src/test/java/com/criteo/publisher/logging/RemoteHandlerTest.kt
```diff
@@ -103,4 +108,17 @@ class RemoteHandlerTest {
 
     verify(sendingQueue).offer(logRecords)
   }
+
+  @Test
+  fun log_GivenConsentNotGiven_DoesNothing() {
+    whenever(consentData.isConsentGiven()).thenReturn(false)
+    val logMessage = LogMessage(message = null)
+    val logRecords = mock<RemoteLogRecords>()
+    whenever(remoteLogRecordsFactory.createLogRecords(logMessage)).thenReturn(logRecords)
```
TurpIF, line 117: When there is no consent, we can even check that the factory is not used at all.
So maybe could you replace this stubbing by a `verify(factory, never()).createLogRecords(any())` ?

publisher-sdk/src/main/java/com/criteo/publisher/logging/RemoteHandler.kt
```diff
@@ -20,24 +20,28 @@ import com.criteo.publisher.SafeRunnable
...
     private val remoteLogRecordsFactory: RemoteLogRecordsFactory,
     private val sendingQueue: ConcurrentSendingQueue<RemoteLogRecords>,
     private val config: Config,
-    private val executor: Executor
+    private val executor: Executor,
+    private val consentData: ConsentData
 ) : LogHandler {
   override fun log(tag: String, logMessage: LogMessage) {
-    fromAndroidLogLevel(logMessage.level)?.takeIf { it >= config.remoteLogLevel } ?: return
+    if (consentData.isConsentGiven()) {
```
TurpIF, line 34: Nit: there is already a lot of `}` at the end of the file. What do you think about writing the condition as a guard instead ?

```
if (!consentGiven) {
    return
}

// Do the job
```

TurpIF (approved):

TurpIF (dismissed):

TurpIF (approved):

PR #194, opened 2021-01-11
Inject deviceId when missing in persisted log records

TurpIF:
Such situation might happen when the log record was created on the main
thread and that the deviceId was not ready yet: for instance for logs
emitted during initialization of the SDK.

The deviceId is assumed to be a constant. Hence, it is not an issue to
set it later. Although, this is quite a workaround based on an
assumption. Hence, the deviceId is kept if it is already set.

JIRA: EE-1411

### DIFF
--- a/test-utils/src/main/java/com/criteo/publisher/csm/ConcurrentSendingQueueHelper.java
+++ b/test-utils/src/main/java/com/criteo/publisher/csm/ConcurrentSendingQueueHelper.java
@@ -16,10 +16,14 @@
 
 package com.criteo.publisher.csm;
 
+import static org.mockito.Mockito.mockingDetails;
+
 public class ConcurrentSendingQueueHelper {
 
   @SuppressWarnings("KotlinInternalInJava")
   public static void emptyQueue(ConcurrentSendingQueue<?> queue) {
-    queue.poll(Integer.MAX_VALUE);
+    if (!mockingDetails(queue).isMock() || mockingDetails(queue).isSpy()) {
+      queue.poll(Integer.MAX_VALUE);
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


test-utils/src/main/java/com/criteo/publisher/csm/ConcurrentSendingQueueHelper.java
```diff
@@ -16,10 +16,14 @@
...
 package com.criteo.publisher.csm;
 
+import static org.mockito.Mockito.mockingDetails;
+
 public class ConcurrentSendingQueueHelper {
 
   @SuppressWarnings("KotlinInternalInJava")
   public static void emptyQueue(ConcurrentSendingQueue<?> queue) {
-    queue.poll(Integer.MAX_VALUE);
+    if (!mockingDetails(queue).isMock() || mockingDetails(queue).isSpy()) {
```
Amokrane, line 25: What's the idea behind verifying that queue isn't a mock? Do you want to override `thenCallRealMethod`?

Amokrane (approved):

test-utils/src/main/java/com/criteo/publisher/csm/ConcurrentSendingQueueHelper.java
TurpIF: The idea is to not add an interaction on a mock. Without this condition, a test was broken because it was doing a `verifyNoInteraction(injectedQueue)`.

A true solution would be to either:
- clean all mock interactions after setup just before the test
- (*) have more finer control on when an action should be done

In this case the method (*) is locally apply to sending queues: it is not necessary to clean it when it is a mock.
I would say that for the moment, something like this is sufficient. If one day we start having issues because of interactions in the setup, then we could think about a solution.

PR #197, opened 2021-01-20
Fix serialization of RemoteLogLevel

TurpIF:
In release artifacts, the enum element was mangled, and even with the
`@SerializedName`, the GSON library still needs to access to enum values
by reflection.

JIRA: EE-1412


vguerci (approved):
Isn't the `SocketException` specialisation unrelated to the PR? Aren't you mixing concerns?

TurpIF:
> Isn't the `SocketException` specialisation unrelated to the PR? Aren't you mixing concerns?

A little bit. When digging about the issue I saw that our serializer was throwing an `Error` rather than an `Exception`. "Thanks" to this, the `Error` reached the publisher code, which alert us. Without this the remote logger would just have been turned off because of silenced exception.

So I restrict this silence to only `SocketException`: we don't care about network errors, but we care about file, serialization, ... errors.

PR #203, opened 2021-02-09
Provide OS (android) to remote configuration

TurpIF:
JIRA: DPP-3177


TurpIF:
@vguerci I didn't added the `deviceModel` field for Android. There is a tons of different models and I think that it would be painful to setup a CaC with it. Anyway we can add it later if really needed.
Although, if you think that I should add it to have the same behavior between ios and Android I'm ok with it.

vguerci (approved):

PR #206, opened 2021-02-23
Explicitly use UTF-8 to write/read JSON

TurpIF:
UTF-8 is generally the default encoding on Android. But it might not be
always the case. For instance, on Windows, the NativeProductTest is
failing without the explicit UTF-8 encoding.


vguerci (commented):
Can you use [`StandardCharsets.UTF_8`](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html#UTF_8)?

vguerci (approved):
Nevermind, approved, apparently this requires API level 19, what's our supported level?

TurpIF:
> Nevermind, approved, apparently this requires API level 19, what's our supported level?

It's Android 16

PR #214, opened 2021-03-01
Stop setting a null webview client

TurpIF:
From documentation: This value cannot be null.

See: https://developer.android.com/reference/android/webkit/WebView#setWebViewClient(android.webkit.WebViewClient)


vguerci (approved):

vguerci:
Sorry I've missed that one! Do not hesitate to ping me on stale reviews like this 😎

PR #215, opened 2021-03-22
Use MavenCentral instead of JCenter to resolve dependencies

leonardlabat:


buildSrc/build.gradle.kts
```diff
@@ -20,9 +20,9 @@ plugins {
 
 repositories {
   google()
-  jcenter()
   gradlePluginPortal()
   maven("https://www.jitpack.io")
+  mavenCentral()
```
TurpIF, line 25: The order of the declarations is the order where Gradle will try to search for the artifacts.
`mavenCentral` will replace `jcenter`, and so represents the main repository (even above `google` I guess).
Could you put it at the top of the declaration please ?

For more info: https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:declaring_multiple_repositories

buildSrc/src/main/java/Repositories.kt
```diff
@@ -30,10 +30,18 @@ fun ScriptHandler.addDefaultInputRepository() {
 
 internal fun RepositoryHandler.addDefaultInputRepository() {
   google()
-  jcenter()
   maven {
     setUrl("https://jitpack.io")
   }
+  jcenter {
```
TurpIF, line 36: Same than in other `build.gradle.kts`: could you move the `mavenCentral` up and leave the `jcenter `where it was before ?

TurpIF (changes requested):

buildSrc/src/main/java/Repositories.kt
```diff
@@ -30,10 +30,18 @@ fun ScriptHandler.addDefaultInputRepository() {
...
+    content {
+      includeGroup("com.mopub")
+      includeGroup("com.criteo.mediation.mopub")
+      includeGroup("com.criteo.mediation.google")
+      includeModule("org.jetbrains.trove4j", "trove4j")
```
TurpIF, line 41: Could you add a link to the Jetbrains YouTrack ? So we can easily check if migration was done.

Also the CI seems red because of detekt relying on kotlinx.html: https://github.com/Kotlin/kotlinx.html/issues/173

buildSrc/src/main/java/Repositories.kt
leonardlabat: done

buildSrc/src/main/java/Repositories.kt
leonardlabat: done

TurpIF (approved):

TurpIF (approved):

PR #216, opened 2021-04-21
Deploy artifacts on Maven Central

TurpIF:
JIRA: DPP-3180

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -21,6 +21,7 @@ plugins {
     id("com.android.application")
     `maven-publish`
     kotlin("android")
+    id("com.vanniktech.android.javadoc") version "0.3.0"
     id("com.jfrog.bintray")
     id("io.gitlab.arturbosch.detekt")
 }
@@ -47,6 +48,12 @@ android {
 
 // Export APK for all build types (release, staging, debug)
 addPublication("Apk") {
+    afterEvaluate {
+        val mainVariant = "memoryLeaksHuntRelease"
+        addSourcesJar(mainVariant)
+        addJavadocJar(mainVariant)
+    }
+
     groupId = "com.criteo.publisher"
     artifactId = "criteo-publisher-sdk-test-app"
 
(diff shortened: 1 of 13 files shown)


app/build.gradle.kts
```diff
@@ -56,24 +63,18 @@ addPublication("Apk") {
     }
 
     android.applicationVariants.all {
+        val isMainVariant = name == mainVariant && false // TODO
```
leonardlabat, line 66: TODO ?

publisher-sdk/src/main/java/com/criteo/publisher/context/UserData.java
```diff
@@ -29,7 +29,7 @@
    * The hashing should be the users’ email address:
    * <ul>
    *   <li>Encoded in UTF-8</li>
-   *   <li>Trimmed of any white space (eg: “test@criteo.com “ should become “test@criteo.com”)</li>
+   *   <li>Trimmed of any white space (eg: “test@criteo.com “ should become “test@criteo.com“)</li>
```
leonardlabat, line 32: Do you want this to be part of that pull request ?

app/build.gradle.kts
TurpIF: Ah, good catch. I was checking if one artifact without classifier was required or not. Answer: all artifacts in the test app have a classifier: debug, staging, release, source, javadoc

publisher-sdk/src/main/java/com/criteo/publisher/context/UserData.java
TurpIF: When pushing to maven local, the javadoc task was complaining about illegal character. I don't know why it is not working locally. But doing a PR dedicated for is kind of overkill. Let me put this in a dedicated commit if it's ok for you.

leonardlabat: Thanks for the explaination !

leonardlabat (approved):

PR #218, opened 2021-04-26
Handle retro compat with Google SDK < 20

TurpIF:
JIRA: DPP-3348


buildSrc/src/main/java/Deps.kt
```diff
@@ -1,21 +1,3 @@
-/*
```
vguerci, line 1: NIT: Any particular reason removing header?

TurpIF:
Coverage is low because the new dedicated module is testing the publisher-sdk module. And coverage across module is not yet supported by the jacoco setup. See https://jira.criteois.com/browse/EE-1370

vguerci (approved):

buildSrc/src/main/java/Deps.kt
TurpIF: Oh, at first sight I thought that my IDE removed this without my consent.
By in fact, you can see that, before, there was 2 headers and that my IDE automatically cleaned one. Smart IDE 😄

vguerci: Smart enough to determine copyright isn't really relevant on deps ¯\\\_(ツ)\_/¯

PR #220, opened 2021-05-06
Bump to 4.4.0

TurpIF:


vguerci (approved):

TurpIF:
There is an issue when downloading the Android emulator. Don't have any details for now. I will enable verbose mode to get more info.

TurpIF:
Seems better now. I do not know if it was fixed because of the bump of Android actions or because Google got a temporary network error.

vguerci (approved):

PR #225, opened 2021-06-29
Support VAST video support for GAM and MoPub AppBidding

TurpIF:
JIRA: DPP-3441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Next Version
+- MoPub AppBidding: Add support for VAST Video ads
+- GAM AppBidding: Add support for VAST Video ads
 
 # Version 4.4.0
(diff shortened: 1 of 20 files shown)


TurpIF:
Note that coverage is low because test-utils coverage report is not generated when run from test module. See https://jira.criteois.com/browse/EE-1370 for more details

leonardlabat (approved):
Just by curiosity : do you know where is the integration guide that explain how to setup the ad srever part ? (because I think that publishers will need to put some conditions there based on the new keyword)

TurpIF:
> Just by curiosity : do you know where is the integration guide that explain how to setup the ad srever part ? (because I think that publishers will need to put some conditions there based on the new keyword)

The public doc is here: https://publisherdocs.criteotilt.com/app/android/app-bidding/google-ad-manager/#ad-server-setup

Currently there is nothing on video as the feature is not released yet

PR #251, opened 2022-02-03
Add support for Android 12

TurpIF:


leonardlabat (dismissed):

buildSrc/src/main/java/AndroidModule.kt
```diff
@@ -94,11 +94,11 @@ fun Project.androidLibModule(configure: AndroidModule.() -> Unit = {}) {
 
 private fun Project.defaultAndroidModule() {
   androidBase {
-    compileSdkVersion(30)
+    compileSdkVersion(31)
```
leonardlabat, line 97: so it doens't need to be 12?

TurpIF: The platform version (12) are the one we're commonly use. But this 31 is the API level which doesn't follow the same incrementation rule. Although you still have almost a 1-1 relationship.
See this table to see both: https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels

PR #260, opened 2022-08-11
🧒 Add coppa support to publisher sdk

viktor-rasevych-criteo:


publisher-sdk/src/main/java/com/criteo/publisher/Criteo.java
```diff
@@ -216,4 +225,6 @@ public void setMopubConsent(@Nullable String ignored) {
 
   public abstract void setUserData(@NonNull UserData userData);
 
+  public abstract void setTagForChildDirectedTreatment(@Nullable Boolean flag);
```
TurpIF, line 228: The wording is quite weird:
* the method is named: `setTagFor...`
* the parameter is named: `flag`

Hence you have 2 different words on the same declaration/meaning. Please use a single one.
Also, I believe that the name `tag for child directed treatment` is a Google word. Do we want to follow Google wording ? Or do we want to use something more generally used in the industry ? For instance:
* IAB: https://www.iab.com/wp-content/uploads/2019/10/IAB_2019-10-09_Navigating-COPPA-Guide.pdf
* Meta: https://developers.facebook.com/docs/audience-network/optimization/best-practices/coppa/

publisher-sdk/src/androidTest/java/com/criteo/publisher/UserPrivacyFunctionalTest.java
```diff
@@ -129,6 +131,36 @@ public void whenCriteoInit_GivenUspOptoutTrue_ThenChangedToFalseAfterFirstCall_V
     assertEquals("false", cdb.getUser().getUspOptout());
   }
 
+  @Test
+  public void whenCriteoInit_GivenTagForChildDirectedTreatmentIsNull_VerifyItNullRegsIsPassedToCdb() throws Exception {
```
TurpIF, line 135: I would not say that the tag is null. It is rather not provided, so as a default value, given no tag, then null is passed to CDB

publisher-sdk/src/test/java/com/criteo/publisher/model/CdbRequestFactoryTest.kt
```diff
@@ -311,6 +313,50 @@ class CdbRequestFactoryTest {
     assertThat(nestedMap).isEqualTo(expectedMap)
   }
 
+  @Test
+  fun createRequest_GivenNotNullTagForChildDirectedTreatment_CreateRequestWithNullCdbRegs() {
```
TurpIF, line 317: GivenNull*, not not null

publisher-sdk/src/test/java/com/criteo/publisher/model/CdbRequestFactoryTest.kt
```diff
@@ -40,6 +40,8 @@ import org.mockito.kotlin.stub
 import org.mockito.kotlin.whenever
 import java.util.concurrent.Future
 import java.util.concurrent.atomic.AtomicInteger
+import kotlin.test.assertEquals
```
TurpIF, line 43: Please use AssertJ/AssertK. It provides lot more helpful error messages

TurpIF (changes requested):

publisher-sdk/src/main/java/com/criteo/publisher/Criteo.java
viktor-rasevych-criteo: We agreed to use Google wording for this method. Will change naming for Boolean parameter

app/src/main/java/com/criteo/testapp/BaseActivity.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import com.criteo.publisher.Criteo
+
+abstract class BaseActivity : AppCompatActivity() {
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    Criteo.getInstance().setTagForChildDirectedTreatment(CoppaFlagHolder.coppaFlag)
```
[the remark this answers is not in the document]
viktor-rasevych-criteo: In this case i wanted to keep the selected TagForChildDirectedTreatment valid after integration reset.
We can simplify things by setting this tag directly on Criteo instance and remove base class. 
For tag preservation after reset we can set old value in new builder

publisher-sdk/src/main/java/com/criteo/publisher/Criteo.java
leonardlabat: @TurpIF PM decision was to re-use Google wording yes

leonardlabat (approved):

TurpIF:
LGTM

TurpIF (commented):
LGTM

TurpIF (approved):

PR #266, opened 2022-09-01
Migrate gson to moshi

viktor-rasevych-criteo:


publisher-sdk/src/main/java/com/criteo/publisher/csm/MetricRequest.kt
```diff
@@ -0,0 +1,84 @@
...
+  constructor(
+      metrics: Collection<Metric>,
+      sdkVersion: String,
+      profileId: Int
+  ) : this(metrics.map { MetricRequestFeedback(it) }, sdkVersion, profileId)
+
+  @OpenForTesting
+  @JsonClass(generateAdapter = true)
+  data class MetricRequestSlot(
+      val impressionId: String?,
```
TurpIF, line 41: Finally, with kotlin null-check, I'm wondering if this field should be non-nullable. Could you try removing the nullability and check if there is any compilation/test errors ?

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+import java.io.IOException
+import java.lang.reflect.Type
+import java.util.Arrays
+
+internal abstract class ArrayListJsonAdapter<C : MutableCollection<T>?, T> private constructor(
```
TurpIF, line 28: why does `C` is nullable ?

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val elementAdapter: JsonAdapter<T>
+) : JsonAdapter<C>() {
+  abstract fun newCollection(): C
+
+  @Throws(IOException::class)
+  override fun fromJson(reader: JsonReader): C {
+    val result = newCollection()
+    reader.beginArray()
+    while (reader.hasNext()) {
+      result?.add(elementAdapter.fromJson(reader)!!)
```
TurpIF, line 38: if `T` is nullable, then it is completely legit to have nullable elements. Then the `!!` will throw an NPE during runtime.

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
```diff
@@ -0,0 +1,108 @@
...
+
+  override fun toString(): String {
+    return "$elementAdapter.collection()"
+  }
+
+  companion object {
+    @JvmStatic
+    val factory = Factory { type, annotations, moshi ->
+      val rawType = Types.getRawType(type)
+      if (annotations.isNotEmpty()) return@Factory null
```
TurpIF, line 61: Could you explain why do you return null for this condition ?

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
```diff
@@ -0,0 +1,108 @@
...
+        type: Type,
+        moshi: Moshi
+    ): JsonAdapter<MutableCollection<T>> {
+      val elementType = Types.collectionElementType(type, Collection::class.java)
+      val elementAdapter: JsonAdapter<T> = moshi.adapter(elementType)
+
+      return object :
+          ArrayListJsonAdapter<MutableCollection<T>, T>(elementAdapter) {
+        override fun newCollection(): MutableCollection<T> {
+          return Arrays.asList()
```
TurpIF, line 88: AFAIK, Arrays.asList is not mutable and this could generate exception at runtime

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    private fun <T> newArrayListAdapter(
+        type: Type,
+        moshi: Moshi
+    ): JsonAdapter<MutableCollection<T>> {
+      val elementType = Types.collectionElementType(type, Collection::class.java)
+      val elementAdapter: JsonAdapter<T> = moshi.adapter(elementType)
+
+      return object :
+          ArrayListJsonAdapter<MutableCollection<T>, T>(elementAdapter) {
```
TurpIF, line 101: ArrayList<T> instead of MutableCollection<T>

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
```diff
@@ -0,0 +1,108 @@
...
+        override fun newCollection(): MutableCollection<T> {
+          return arrayListOf()
```
TurpIF, line 103: Why not just using the Java constructor directly ?

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/BooleanJsonAdapter.kt
```diff
@@ -0,0 +1,44 @@
...
+
+package com.criteo.publisher.util.jsonadapter
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonDataException
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import java.lang.Boolean.parseBoolean
+
+internal class BooleanJsonAdapter : JsonAdapter<Boolean>() {
```
TurpIF, line 25: Could you add a comment explaining what this adapter is doing and why

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
```diff
@@ -0,0 +1,108 @@
...
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+import java.io.IOException
+import java.lang.reflect.Type
+import java.util.Arrays
+
```
TurpIF, line 27: Could you add a comment explaining what this adapter is doing and why

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/BooleanJsonAdapter.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  override fun toJson(writer: JsonWriter, value: Boolean?) {
+    if (value == null) {
+      throw NullPointerException(
+          "value was null! Wrap in .nullSafe() to write nullable values.")
+    }
+    writer.value(value)
+  }
```
TurpIF, line 43: Shouldn't you override the toString as in other adapters ?

publisher-sdk/src/test/java/com/criteo/publisher/csm/MetricParserTest.kt
```diff
@@ -1,179 +0,0 @@
...
-import org.mockito.kotlin.verify
-import java.io.ByteArrayInputStream
-import java.io.ByteArrayOutputStream
-import java.io.EOFException
-import java.io.IOException
-import java.io.InputStream
-import java.io.OutputStream
-import javax.inject.Inject
-
-class MetricParserTest {
```
TurpIF, line 38: Do you mind checking if there is test that are worth keeping ? Notably on the Metric serialization

publisher-sdk/src/test/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapterTest.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  @Rule
+  @JvmField
+  var mockedDependenciesRule = MockedDependenciesRule()
+
+  @Inject
+  private lateinit var moshi: Moshi
+
+  @Test
+  fun givenArraysArrayList_toJson_ShouldSerializeCorrectly() {
```
TurpIF, line 37: Those classes are really technical and could deserve a better test coverage. Could you add more tests for those scenarios:
- what about deserialization ? for both ArrayList and Arrays.ArrayList ?
- what if there is null elements in the lists (a `List<T?>`)
- what if the list is null ?
- what if the list is empty ?
- what if the list contains a sublist ?
- what if the list as different kind of elements ? like a `List<Any>` such as `listOf(1, true, "foo")` ?

publisher-sdk/src/test/java/com/criteo/publisher/util/jsonadapter/BooleanJsonAdapterTest.kt
```diff
@@ -0,0 +1,70 @@
...
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.criteo.publisher.util.jsonadapter
+
+import com.squareup.moshi.JsonDataException
+import org.assertj.core.api.Assertions
```
TurpIF, line 20: you could imports the static methods from Assertions

publisher-sdk/src/test/java/com/criteo/publisher/util/jsonadapter/BooleanJsonAdapterTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Before
+import org.junit.Test
+
+class BooleanJsonAdapterTest {
+
+  private lateinit var adapter: BooleanJsonAdapter
```
TurpIF, line 26: I would rather test from the provided mochi object rather than on the adapter itself. For sure testing the adapter directly is more unitary, but from SDK point of view, mochi (with the adapter) is used, not the adapter itself

publisher-sdk/src/test/java/com/criteo/publisher/util/jsonadapter/URIAdapterTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test
+import java.net.URI
+
+class URIAdapterTest {
+
+  private lateinit var adapter: JsonAdapter<URI>
+
+  @Before
+  fun setUp() {
+    adapter = URIAdapter().lenient()
```
TurpIF, line 33: Same, I would rather test again the mochi object.

publisher-sdk/src/test/java/com/criteo/publisher/util/jsonadapter/URLAdapterTest.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.junit.Before
+import org.junit.Test
+import java.net.URL
+
+class URLAdapterTest {
+  private lateinit var adapter: JsonAdapter<URL>
+
+  @Before
+  fun setUp() {
+    adapter = URLAdapter().lenient()
```
TurpIF, line 31: Same, I would rather test again the mochi object.

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/URIAdapter.kt
```diff
@@ -0,0 +1,52 @@
...
+package com.criteo.publisher.util.jsonadapter
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonDataException
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import java.io.IOException
+import java.net.URI
+
+internal class URIAdapter : JsonAdapter<URI>() {
```
TurpIF, line 26: From https://github.com/square/moshi#custom-type-adapters, it is said:
> With Moshi, it’s particularly easy to customize how values are converted to and from JSON. A type adapter is any class that has methods annotated @ToJson and @FromJson.

> Note that the method annotated with @FromJson does not need to take a String as an argument. Rather it can take input of any type and Moshi will first parse the JSON to an object of that type and then use the @FromJson method to produce the desired final value. Conversely, the method annotated with @ToJson does not have to produce a String.

Hence I'm wondering if you could ease this adapter by just having a mapper of String <-> URI

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/URLAdapter.kt
```diff
@@ -0,0 +1,52 @@
...
+package com.criteo.publisher.util.jsonadapter
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonDataException
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import java.io.IOException
+import java.net.URL
+
+internal class URLAdapter : JsonAdapter<URL>() {
```
TurpIF, line 26: Same than for URI adapter, maybe you could just have a mapper of String <-> URL

publisher-sdk/src/main/java/com/criteo/publisher/DependencyProvider.java
```diff
@@ -598,30 +603,31 @@ public MetricSendingQueueConfiguration provideMetricSendingQueueConfiguration()
...
-    return getOrCreate(Gson.class, () -> new GsonBuilder()
-        .registerTypeAdapterFactory(CustomAdapterFactory.create())
-        .create());
+  public Moshi provideMoshi() {
+    return getOrCreate(Moshi.class, () -> new Moshi.Builder()
+        .add(
+            RemoteLogLevel.class,
+            EnumJsonAdapter.create(RemoteLogLevel.class).withUnknownFallback(null).nullSafe()
+        )
+        .add(URI.class, new URIAdapter().lenient())
```
TurpIF, line 625: Why do you need lenient adapters ?

publisher-sdk/src/main/java/com/criteo/publisher/DependencyProvider.java
```diff
@@ -598,30 +603,31 @@ public MetricSendingQueueConfiguration provideMetricSendingQueueConfiguration()
...
+        .add(URL.class, new URLAdapter().lenient())
+        .add(Boolean.class, new BooleanJsonAdapter().lenient().nullSafe())
+        .add(boolean.class, new BooleanJsonAdapter().lenient().nullSafe())
+        .add(ArrayListJsonAdapter.getFactory())
```
TurpIF, line 629: I'm very curious why you need a so specific collection adapter, isn't mochi already handling lists ?

publisher-sdk/src/main/java/com/criteo/publisher/csm/MetricRequest.kt
viktor-rasevych-criteo: Sure. Just tested, no compilation and no unit test errors

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
viktor-rasevych-criteo: Yep, it should not be nullable. This class is taken from moshi library itself (from [here](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/CollectionJsonAdapter.kt)), but it is internal, so we cannot just add additional factory. Will synchronize it with moshi version.

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
viktor-rasevych-criteo: No longer nullable. Synchronized code with moshi version. See comment above

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
viktor-rasevych-criteo: This class is taken from moshi code see the same condition [here](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/CollectionJsonAdapter.kt#L51)

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
viktor-rasevych-criteo: Yep, you are right. Actually we need this class only for object to json conversion. At runtime when we get moshi adapter it is fetched for a runtime class. This can be fixed by converting serialization code to kotlin and using inline function with reified generic. But doing so requires a lot of changes in code and we will need to get moshi statically from inline function.

I will write some comments on when this class is used

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/ArrayListJsonAdapter.kt
viktor-rasevych-criteo: Yep, will change that

publisher-sdk/src/main/java/com/criteo/publisher/util/jsonadapter/BooleanJsonAdapter.kt
viktor-rasevych-criteo: As far i see from tests, sometimes boolean value comes as string. For example [here](https://github.com/criteo/android-publisher-sdk/blob/3eed94efce7629818211f8a890c8d5d7bef64965/publisher-sdk/src/test/java/com/criteo/publisher/model/CdbResponseSlotTest.java#L579). To not break this logic i added this adapter.

viktor-rasevych-criteo:
Removed ArrayListJsonAdapter and handled List parsing inside JsonSerializer.

leonardlabat (approved):

PR #267, opened 2022-09-16
⬆️ Add support for Android 12L (API 32)

viktor-rasevych-criteo:
JIRA: DPP-3857

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -90,7 +90,11 @@ jobs:
               tag: 'playstore' # Explicitly need the PlayStore to fetch GAID
 
             - api: '31'
-              abi: 'x86_64' # x86 ABI is not yet available for API 31
+              abi: 'x86_64'
+              tag: 'google_apis'
+
+            - api: '32'
+              abi: 'x86_64'
               tag: 'google_apis'
     steps:
       - name: Checkout (latest)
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -92,6 +92,10 @@ jobs:
             - api: '31'
               abi: 'x86_64' # x86 ABI is not yet available for API 31
               tag: 'google_apis'
+
+            - api: '32'
+              abi: 'x86_64'
```
TurpIF, line 97: Do you use the x86_64 ABI as it is not made available by Google ? (yet?) 
Could you please check if x86 is available on API 31 ? If not then maybe we can assume that on API >= 31, then only x86_64 is available and would could move/rewrite the comment above ?

viktor-rasevych-criteo: Yes, x86 is not available for API 31. The same is true for API 32 and 33. I guess we can assume Google won't be releasing x86 emulators. I will remove comment then

viktor-rasevych-criteo: image

viktor-rasevych-criteo: It is from [here](https://developer.android.com/about/versions/12/get)

leonardlabat (approved):

PR #269, opened 2022-09-26
🔖 Bump to version 4.8.0

viktor-rasevych-criteo:


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Next Version
+
+# Version 4.8.0
 - Add support of Android 12L (API level 32)
 - Bump AdMob to v21.2.0
+- Bump minSdkVersion to 19
```
TurpIF, line 6: By doing that you can remove all the degraded mode of the SDK:
- https://github.com/criteo/android-publisher-sdk/blob/main/publisher-sdk/src/main/java/com/criteo/publisher/util/DeviceUtil.java#L65 become true and could be inlined everywhere, then removed
- https://github.com/criteo/android-publisher-sdk/blob/main/publisher-sdk/src/main/java/com/criteo/publisher/DummyCriteo.java could be removed
- CriteoInternal class could be inlined in the Criteo abstract class
- Many tests about degraded features could be removed

publisher-sdk/src/test/java/com/criteo/publisher/advancednative/AdChoiceOverlayNativeRendererTest.kt
```diff
@@ -53,6 +58,8 @@ class AdChoiceOverlayNativeRendererTest {
   @Before
   fun setUp() {
     renderer = AdChoiceOverlayNativeRenderer(delegate, adChoiceOverlay)
+    // Main Lopper is null for unit tests, so just mock executor to avoid NullPointerException
+    whenever(mockedDependenciesRule.dependencyProvider.provideRunOnUiThreadExecutor()).thenReturn(mock())
```
TurpIF, line 62: This is true for all Java tests. Instead of having a particular mock on this test, what about setting this directly in the MockedDependenciesRule so that a java test only needs to setup it to have all env mocked correctly.

For instance, for Application, the rule is setting up a mock Application via this class: https://github.com/criteo/android-publisher-sdk/blob/main/test-utils/src/main/java/com/criteo/publisher/application/ApplicationResource.java
Only for java tests: https://github.com/criteo/android-publisher-sdk/blob/main/test-utils/src/main/java/com/criteo/publisher/application/InstrumentationUtil.java#L33

For the UI thread executor, maybe could you mock it the same way by extending/reusing this component: https://github.com/criteo/android-publisher-sdk/blob/main/test-utils/src/main/java/com/criteo/publisher/concurrent/MultiThreadResource.java

CHANGELOG.md
viktor-rasevych-criteo: Let's keep the degraded mode to simplify API deprecation in future

TurpIF: LGTM

viktor-rasevych-criteo: Done

test-utils/src/main/java/com/criteo/publisher/concurrent/MultiThreadResource.java
```diff
@@ -54,6 +56,10 @@ public void setUp() {
 
     dependencyProvider.inject(Executor.class, trackingCommandsExecutor);
     dependencyProvider.inject(AsyncResources.class, trackingCommandsExecutor.asAsyncResources());
+    if (!InstrumentationUtil.isRunningInInstrumentationTest()) {
+      // Main Lopper is not available in unit tests, so just mock RunOnUiThreadExecutor to avoid NullPointerException
+      dependencyProvider.inject(RunOnUiThreadExecutor.class, mock(RunOnUiThreadExecutor.class));
```
TurpIF, line 61: You might want to have a real executor rather than a mock that will do nothing

TurpIF: As UI thread is a unique thread, you could use the single thread executor from java (https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newSingleThreadExecutor()) and wrap it inside a TrackingCommandsExecutor which would be waited for during the tear down

viktor-rasevych-criteo: This is a bit more complicated to do since we need an instance of RunOnUiThreadExecutor. We could subclass RunOnUiThreadExecutor with class that has TrackingCommandsExecutor in constuctor and delegate all method calls

viktor-rasevych-criteo: Alternative approarch is to use `DirectMockRunOnUiThreadExecutor` which will run code on thread we currently on.
Using separate thread adds some flakiness to unit tests

TurpIF (approved):

test-utils/src/main/java/com/criteo/publisher/concurrent/MultiThreadResource.java
TurpIF: Ok

PR #271, opened 2022-10-05
Obfuscated code crash

viktor-rasevych-criteo:


publisher-sdk/src/main/java/com/criteo/publisher/logging/RemoteLogRecords.kt
```diff
@@ -46,6 +46,7 @@ data class RemoteLogRecords(
       @Json(name = "deviceOs") val deviceOs: String?
   )
 
+  @JsonClass(generateAdapter = false)
```
TurpIF, line 49: all other occurrence of `@JsonClass` are setting `generateAdapter=true`, while here you're setting it to `false`.
Could you confirm the `false` value ?

Also, I see no automated tests in this PR. Is there anyway to automatically mitigate such kind of error ?
If automation is hard or not possible, then I advise to add a new item in the manual bugfest until the automation is developed.

viktor-rasevych-criteo: Yes, for this case we use `false` since we are using manually written adapter from Moshi lib -> [EnumJsonAdapter](https://github.com/square/moshi/blob/master/moshi-adapters/src/main/java/com/squareup/moshi/adapters/EnumJsonAdapter.kt). The reason why we need such annotation here is for the purpose of properly applied [proguard](https://github.com/square/moshi/blob/master/moshi/src/main/resources/META-INF/proguard/moshi.pro) rules from Moshi.

I guess we need some automation for shrinked/obfuscated artifacts to avoid such errors in future. Running tests against such artifacts should do the trick.

TurpIF: ok, thank you for the explanation

TurpIF (approved):

PR #274, opened 2022-12-14
Inject mraid into webview

viktor-rasevych-criteo:


publisher-sdk/src/main/java/com/criteo/publisher/adview/AdWebViewClient.kt
```diff
@@ -61,7 +63,8 @@ class AdWebViewClient(
 
         WebResourceResponse("text/javascript", "UTF-8", stream)
       } catch (e: IOException) {
-        throw RuntimeException(e)
+        logger.log(MraidLogMessage.onErrorDuringMraidFileInject(e))
```
TurpIF, line 66: The last commits is overriding some things that were done in the "Inject mraid scrpt" commit like if this previous commit was simply a WIP.

Maybe it is better to squash them together ? You need the test to fixup the implementation, so both test and implementations are coming together naturally.

Also, on a general note, do you mind doing the transformation from Java to Kotlin in a dedicated commit (ie: one commit for the renaming to keep git history + 1 commit for the actual migration + 1 commit for your added feature) ? So it will ease the review and the MRAID commit will be much smaller.

viktor-rasevych-criteo: Sure, will cleanup my commits!

viktor-rasevych-criteo: Updated

publisher-sdk/src/androidTest/java/com/criteo/publisher/adview/AdWebViewClientTest.kt
```diff
@@ -52,28 +59,37 @@ class AdWebViewClientTest {
...
   private lateinit var context: Context
 
+  @MockBean
+  private lateinit var logger: Logger
+
   @Mock
   private lateinit var listener: RedirectionListener
   private lateinit var webView: WebView
   private lateinit var webViewClient: AdWebViewClient
+  private val pageFinishedLoading = CompletableFuture<Unit>()
```
TurpIF, line 78: nit: Such logic are generally synchronized with [CountDownLatch](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html) objects. But I'm not sure if they are supported by Android 🤷‍♂️

TurpIF (commented):
LGTM

publisher-sdk/src/androidTest/java/com/criteo/publisher/adview/AdWebViewClientTest.kt
viktor-rasevych-criteo: I used [this](https://github.com/criteo/android-publisher-sdk/blob/main/publisher-sdk/src/main/java/com/criteo/publisher/util/CompletableFuture.java) implementation of CompletableFuture which uses CountDownLatch under the hood. We actually have android version of CountDownLatch and it can be used here

afewcc (approved):

PR #275, opened 2023-01-16
💪 Add SDK report about MRAID readiness

viktor-rasevych-criteo:
- Report MRAID readiness
- Add ability to report error on MRAID object from SDK

JIRA: DPP-4237

### DIFF
--- a/publisher-sdk/src/androidTest/java/com/criteo/publisher/adview/AdWebViewClientTest.kt
+++ b/publisher-sdk/src/androidTest/java/com/criteo/publisher/adview/AdWebViewClientTest.kt
@@ -20,14 +20,15 @@ import android.content.Context
 import android.webkit.WebView
 import androidx.test.rule.ActivityTestRule
 import com.criteo.publisher.MraidData
+import com.criteo.publisher.callMraidObjectBlocking
 import com.criteo.publisher.concurrent.ThreadingUtil
 import com.criteo.publisher.concurrent.ThreadingUtil.runOnMainThreadAndWait
+import com.criteo.publisher.loadMraidHtml
 import com.criteo.publisher.logging.Logger
 import com.criteo.publisher.mock.MockBean
 import com.criteo.publisher.mock.MockedDependenciesRule
 import com.criteo.publisher.mock.SpyBean
 import com.criteo.publisher.test.activity.DummyActivity
-import com.criteo.publisher.util.CompletableFuture
 import com.criteo.publisher.view.WebViewClicker
 import com.criteo.publisher.view.WebViewLookup
 import org.assertj.core.api.Assertions.assertThat
@@ -36,13 +37,13 @@ import org.junit.Rule
 import org.junit.Test
 import org.mockito.ArgumentMatchers
 import org.mockito.Mock
+import org.mockito.junit.MockitoJUnit
 import org.mockito.kotlin.inOrder
 import org.mockito.kotlin.never
 import org.mockito.kotlin.spy
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.verifyNoMoreInteractions
 import org.mockito.kotlin.whenever
-import org.mockito.junit.MockitoJUnit
 import java.io.IOException
 
 class AdWebViewClientTest {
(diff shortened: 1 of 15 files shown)


publisher-sdk/src/androidTest/java/com/criteo/publisher/adview/AdWebViewClientTest.kt
```diff
@@ -186,6 +189,28 @@ class AdWebViewClientTest {
...
   }
 
+  @Test
+  fun whenHtmlWithMraidIsLoaded_GivenPageFinished_ShouldHaveMraidStateAsDefault() {
+    val mraidHtml = mraidData.getHtmlWithMraidScriptTag()
+
+    loadHtml(mraidHtml)
+    waitForPageToFinishLoading()
+
+    verifyMraidObject(".getState()") { state ->
```
TurpIF, line 199: It is kind of feeling weird to see a `".getState()"` like this. Why not having `"mraid"` as default value of the method so that other calls could looks like `"mraid.getState()"`

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
```diff
@@ -0,0 +1,37 @@
...
+@OpenForTesting
+internal class MraidInteractor(private val webView: WebView) {
+
+  fun notifyReady() {
+    callOnMraidObject("notifyReady()")
+  }
+
+  fun notifyError(message: String, action: String? = null) {
+    val actionString = if (action == null) "" else ", \"$action\""
+    callOnMraidObject("notifyError(\"$message\"$actionString)")
```
TurpIF, line 31: For now you have a single JS call, but maybe when you have plenty of those, you could think about a utility method to transform a list of args into JS args. Like:
- given a string `foo` -> `"foo"`
- given an int `42` -> `42`
- given a null -> `undefined`

And joining the args with comma. So you won't have to deal (and maybe forget) the quotes around strings or handling the nulls

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  private fun callOnMraidObject(jsCode: String) {
+    webView.evaluateJavascript("window.mraid.$jsCode", null)
```
TurpIF, line 35: Another option is to log debug all calls so that you could clearly see if there is any error when interacting with MRAID.

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
viktor-rasevych-criteo: Great idea, updated my code

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
viktor-rasevych-criteo: Added call log to simplify debugging in case of problem

publisher-sdk/src/androidTest/java/com/criteo/publisher/adview/AdWebViewClientTest.kt
viktor-rasevych-criteo: Updated code, so now it is just `"getState()"`. The same way it is used in `MraidInteractor`

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
```diff
@@ -0,0 +1,51 @@
...
+internal class MraidInteractor(private val webView: WebView) {
+
+  private val logger = LoggerFactory.getLogger(MraidInteractor::class.java)
+
+  fun notifyReady() {
+    callOnMraidObject("notifyReady()")
+  }
+
+  fun notifyError(message: String, action: String? = null) {
+    callOnMraidObject("notifyError(${asJsArgs(message, action)})")
```
TurpIF, line 33: nit: you could use suffix functions on string to get something like:

```
"notifyError".jsCall(message, action)
```

Then `jsCall` would to the same as your `asJsArgs` but add also the parenthesis.
Or it could even call directly the `callOnMraidObject` ?

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
```diff
@@ -0,0 +1,51 @@
...
+    logger.debug("Calling mraid object with js: $jsToEvaluate")
+    webView.evaluateJavascript(jsToEvaluate, null)
+  }
+
+  private fun asJsArgs(vararg params: Any?): String {
+    return params.joinToString(separator = ", ") {
+      when (it) {
+        null -> "undefined"
+        is String -> "\"$it\""
+        else -> it.toString()
```
TurpIF, line 47: I would rather throw in default case. Normally, if you need a new type, you should quickly detect it with the exception and fix it. It will prevent you to blindly guess that all types are supported while they are not

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
viktor-rasevych-criteo: What do you think about making it invoke operator. It will look like this 
```
"notifyError"(message, action)
```

TurpIF: LGTM

publisher-sdk/src/main/java/com/criteo/publisher/adview/MraidInteractor.kt
viktor-rasevych-criteo: Okay, it makes sense in a long run. Will do the change

afewcc (approved):

PR #276, opened 2023-01-18
Update to android 13

viktor-rasevych-criteo:


.github/workflows/main.yml
```diff
@@ -93,7 +93,7 @@ jobs:
               abi: 'x86_64'
               tag: 'google_apis'
 
-            - api: '32'
+            - api: '33'
```
TurpIF, line 96: you should add a new version to support, not replace one.

TurpIF: It might be worth to add a comment explaining why it is skipped then.

viktor-rasevych-criteo: Added comment

afewcc (approved):

PR #279, opened 2023-02-15
Updated test app to use demo placements

leonardlabat:


leonardlabat:
@TurpIF do you think about any reason why we'd like to keep the cdb mock case for the test app rather than always using the prod cdb backend?

viktor-rasevych-criteo (approved):

PR #288, opened 2023-06-12
Fix flaky test

viktor-rasevych-criteo:


.github/workflows/main.yml
```diff
@@ -146,24 +144,6 @@ jobs:
             adb logcat >> artifacts/logcat.log &
             ./gradlew gordon --tests=androidx.test.filters.FlakyTest && ./gradlew connectedCheck -Pandroid.testInstrumentationRunnerArguments.notAnnotation=androidx.test.filters.FlakyTest
```
TurpIF, line 145: Just note that using Gordon will have some cost:
- execution time is generally slower. We can even notice it on this PR ([~1h5min](https://github.com/criteo/android-publisher-sdk/actions/runs/5246415431)) vs a previous one ([~45min](https://github.com/criteo/android-publisher-sdk/actions/runs/5188687665))
- named parameterized tests are silently ignored. See this issue: https://github.com/Banno/Gordon/discussions/47.

While the first point can argued as it might be better to pay one time 1h than multiple time 45 minutes, I guess the second point is quite important as it means that some tests are simply not run at all.

Indeed, with this PR, we expect the number of test to grow, but older PR shows that 554 tests were run ([report](https://github.com/criteo/android-publisher-sdk/actions/runs/5188687665/jobs/9361366848#step:8:2074)), and there is still 554 tests in this PR too ([report](https://github.com/criteo/android-publisher-sdk/actions/runs/5246415431/jobs/9478838782?pr=288#step:7:4533)).

Do you mind enabling the JUnit XML export temporary to confirm that no tests are ignored ?

TurpIF (changes requested):

.github/workflows/main.yml
viktor-rasevych-criteo: From what i can see there are some ignored tests, like [here](https://github.com/criteo/android-publisher-sdk/actions/runs/5246415431/jobs/9478838620?pr=288) Those all are from `DfpHeaderBiddingFunctionalTest`. I checked a bit older action which runs all the test and those tests were always ignored. 
[Here](https://github.com/criteo/android-publisher-sdk/actions/runs/5246415431/jobs/9478838620?pr=288#step:7:4366) you can find a gordon report. It reports that 22 tests are ignored (that's all the tests from `DfpHeaderBiddingFunctionalTest`)

TurpIF: I guess the named parameter here should be removed: https://github.com/criteo/android-publisher-sdk/blob/a398ec6bd24e0fb8e8ee3fd1fbe4da4a5bacf0f2/publisher-sdk/src/androidTest/java/com/criteo/publisher/integration/DfpHeaderBiddingFunctionalTest.java#L138

Those tests where already flagged as `@FlakyTest`, so it is not a regression.

But I guess the InHouse functional tests (which where not flagged as flaky) will then be ignored too. See: https://github.com/search?q=repo%3Acriteo%2Fandroid-publisher-sdk%20%40Parameters(name&type=code

TurpIF: Oups, my bad. Gordon will still only applied to `@FlakyTest`.
The DFP tests could be enabled in another PR I guess still.

TurpIF (approved):

PR #293, opened 2023-07-14
❔ Send SDK Mraid capability in bid request

viktor-rasevych-criteo:
JIRA: DPP-4565

### DIFF
--- a/publisher-sdk/src/main/java/com/criteo/publisher/DependencyProvider.java
+++ b/publisher-sdk/src/main/java/com/criteo/publisher/DependencyProvider.java
@@ -355,7 +355,8 @@ public class DependencyProvider {
         provideBuildConfigWrapper(),
         provideIntegrationRegistry(),
         provideContextProvider(),
-        provideUserDataHolder()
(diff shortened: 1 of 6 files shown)


publisher-sdk/src/main/java/com/criteo/publisher/model/SdkFeature.kt
```diff
@@ -0,0 +1,27 @@
...
+ *    limitations under the License.
+ */
+
+package com.criteo.publisher.model
+
+/**
+ * Contains codes for features of SDK described in OpenRTB Api Spec
+ */
+@Suppress("MagicNumber")
+enum class SdkFeature(val code: Int) {
```
leonardlabat, line 23: Do you mind renaming to ApiFramework?
That's the name of the enum in ortb specficiation

viktor-rasevych-criteo: sure

leonardlabat (approved):

